PR #5, opened 2020-04-16
Simplify substratevm built

zakkak:
Tested with `mx build --projects SVM_HOSTED`
Note that this won't give you an `svm.jar` but an `svm-hosted.jar`, which I guess you can rename...

### DIFF
--- a/src/build.java
+++ b/src/build.java
@@ -164,40 +164,9 @@ class Mx
     private static final Logger LOG = LogManager.getLogger(OperatingSystem.class);
     private static final Pattern VERSION_PATTERN = Pattern.compile("\"([0-9]\\.[0-9]{1,3}\\.[0-9]{1,2})\"");
 
-    private static final String SVM_ONLY = String.join(","
-        , "SVM"
-        , "com.oracle.svm.graal"
-        , "com.oracle.svm.truffle"
-        , "com.oracle.svm.hosted"
-        , "com.oracle.svm.truffle.nfi"
-        , "com.oracle.svm.truffle.nfi.posix"
-        , "com.oracle.svm.truffle.nfi.windows"
-        , "com.oracle.svm.core.jdk11"
-        , "com.oracle.svm.core"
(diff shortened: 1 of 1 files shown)


galderz:
Thanks @zakkak, I'll try it out locally shortly.

galderz:
@zakkak Tried it and it works, and it does seem to give you an `svm.jar`. This is located in `/tmp/mandrel/substratevm/mxbuild/dists/jdk11/svm.jar` when I run it locally inside the docker container.

I've double-checked and this jar has indeed the same contents as the normal `svm.jar`. And of course, no libffi is downloaded.

Bravo 👏👏👏

galderz:
Oh one sec, that jar might be from another run (since volumes are use)

zakkak:
Yes that's probably from a different run.
I think that's easy to check by inspecting the modification timestamps of the jars ;)

galderz:
@zakkak I've pushed some changes to master to do an `mx clean` before the build to highlight errors introduced by changes like this.

So, jar name aside, there are some clear differences between the two jars (see [here](https://gist.github.com/galderz/236f47e7bf3faaced3b96b335d8e3ee8)). The files there are missing from `svm-hosted.jar` as opposed to `svm.jar`.

For this change to be acceptable, you should provide some proof that Quarkus works fine without these classes/files. Also, ideally we'll produce a single jar that works for the Maven dependency in Quarkus JVM mode, and for the native image generation in Quarkus Native mode. So, the verification should really apply to both. Have you already carried out such test? If not, could you look into it?

zakkak:
Hmm unfortunately it doesn't work. The first issue I came across is a dependency on `com.oracle.svm.core.posix` which is not included in the `svm-hosted.jar`.

And I see that this package is only included in the SVM _mx project_ which brings the libffi as well :(

galderz:
Pity. Thanks for giving it a go though @zakkak!

PR #6, opened 2020-04-21
Remove ignored --dependencies

zakkak:
--dependencies/--projects are ignored when combined with --only
See https://github.com/graalvm/mx/issues/211


zakkak:
Ignore... I misinterpreted the code.

PR #14, opened 2020-05-27
Update for GraalVM 20.1

galderz:
Update and some refactoring to simplify things.

More unit tests will be added in the near future, now that external dependencies are separated properly. There's only 1 unit test now, which runs before executing anything (it's fast).


zakkak:
LGTM, didn't go through all the changes in `src/build.java` though.

PR #15, opened 2020-05-27
Adds ansible playbook for setting up packaging environment

zakkak:
Supports:
* remote hosts
* local containers
* multiple configurations


ansible/playbook.yml
```diff
@@ -0,0 +1,38 @@
+---
+- name: Setup machine for mandrel-packaging
+  hosts: all
+  vars:
+    configuration: mandrel20.1.0-openjdk
+    mx_path: /opt/mx
+    mandrel_path: /tmp/mandrel
+    mandrel_repo: https://github.com/oracle/graal.git
```
galderz, line 8: This should be now https://github.com/graalvm/mandrel ?

galderz:
@zakkak I'm giving this a go locally too :)

ansible/playbook.yml
zakkak: Good catch, I started this before settling down on graalvm/mandrel :)

Fix on the way (testing it).

zakkak:
Hi @galderz 
I think I tackled most (if not all of your comments).

* removed mandrel user.
* altered Makefile to use ansible (simply running `make` should take care of everything and give you a prompt in a container.
* When you open a shell in the container you get a message on how to build.
* Updated Readme and added instructions on how to copy local files to the container when developing.

galderz:
Hey Foivos, thanks for this, but not sure things are there yet:

* `make` fails if nothing running, so I made stop/rm idempotent (see [PR](https://github.com/zakkak/mandrel-packaging/pull/2)).
* After fixing that run `make` but don't see the welcome message you prepared (see [output](https://gist.github.com/galderz/d8af56ac5a96df019ae04245874717f1)). How/when does the welcome message appear?
* I tried the welcome message example but doesn't work. `java` is not in the classpath. The message seems to imply you'd run this from `/opt/jdk/bin` which is odd. I'd expect that either you run it from `/` or `/root`.
* After going into `/opt/jdk/bin` I tried the `./java ...` command and it failed with [this](https://gist.github.com/galderz/9afdbcb47f7bd10e12920220662ebd3a). I already fixed this in the PR I sent you last week :\ (see 2nd and 3rd fixes in https://github.com/zakkak/mandrel-packaging/pull/1). Did you try this?

galderz:
Also, I'm unsure about the need to copy any source files across to the container. A volume works just as well and there's no copying required. It just establishes a link that makes iterative development much easier to do.

zakkak:
> Hey Foivos, thanks for this, but not sure things are there yet:
> 
>     * `make` fails if nothing running, so I made stop/rm idempotent (see [PR](https://github.com/zakkak/mandrel-packaging/pull/2)).
> 

Looks like I had something running everytime I tried it. Good catch. Thanks

>     * After fixing that run `make` but don't see the welcome message you prepared (see [output](https://gist.github.com/galderz/d8af56ac5a96df019ae04245874717f1)). How/when does the welcome message appear?
> 

According to the log the container didn't `stop-image` before `start-image-attach`.
If the container is already started the message is not shown on attach.
I will have a look to see if this can be fixed with different `-i -t -d` options.

>     * I tried the welcome message example but doesn't work. `java` is not in the classpath. The message seems to imply you'd run this from `/opt/jdk/bin` which is odd. I'd expect that either you run it from `/` or `/root`.
> 

Oh sorry for that. The `./` part was a leftover it should be `java ...` fixed

>     * After going into `/opt/jdk/bin` I tried the `./java ...` command and it failed with [this](https://gist.github.com/galderz/9afdbcb47f7bd10e12920220662ebd3a). I already fixed this in the PR I sent you last week :\ (see 2nd and 3rd fixes in [zakkak#1](https://github.com/zakkak/mandrel-packaging/pull/1)). Did you try this?

Yes, and I cherry-picked it in https://github.com/mandrel/mandrel-packaging/pull/17 where I use it, but it's no longer needed for this PR.

zakkak:
> Also, I'm unsure about the need to copy any source files across to the container. A volume works just as well and there's no copying required. It just establishes a link that makes iterative development much easier to do.

I know but AFAIK there is no way to dynamically attach a volume, so I think that we need to make the  `make boot-fedora-image` parametrizable to make this work. I would prefer to do this in a separate PR if you don't mind.

galderz:
Trying the latest changes...

galderz:
@zakkak I've sent a PR for a small fix about the `run-image` target.

The build is unable to complete, when calling:

```
$ java -ea /root/src/build.java --install --version 20.1.0.redhat-00001 --verbose
```

You get (see [longer output](https://gist.github.com/galderz/b6787bca59062b102474a29c2afd0f7f)):

```
Caused by: java.nio.file.NoSuchFileException: /resources/assembly/graal-sdk/pom.xml
```

Seems like you only copied over the src/build.java? You need the rest of metadata too to build the Maven artifacts correctly.

I think the metadata should be alongside `src`, like in the repo itself.

Then, the script expects the metadata to be somewhere relative to the script, so I'd do also the following:

* Make the folder when `make run-image` to be `/root`.
* Adjust the messages shown so that the examples are relative to that, e.g.

```
  java -ea src/build.java --install --version 20.1.0.redhat-00001 --verbose
```

galderz:
Needs rebase.

galderz:
Once this is in place, it'd be nice to enhance `build.java` maybe with a parameter/command that allows the extra step to build the JDK to be executed. It'd essentially call the existing script, with the right paths and would run inside the container. @zakkak WDYT?

zakkak:
@galderz thanks for testing

```
make build-image
make run-image
java -ea src/build.java --install --version 20.1.0.redhat-00001
```
are working now.

> Once this is in place, it'd be nice to enhance `build.java` maybe with a parameter/command that allows the extra step to build the JDK to be executed. 

By "the JDK to be executed" you refer to the JDK patched with the Mandrel bits (what I like to call MandrelJDK).

> It'd essentially call the existing script, with the right paths and would run inside the container. @zakkak WDYT?

The existing script being buildJDK?

@jerboaa is looking into how to integrate his build scripts with ours to avoid duplicating efforts.
So maybe openning an issue to further discuss this is better than doing it here. 
It also sounds a bit unrelated to the ansible playbook.

zakkak:
Pushed some more changes so now you can run:

```
make build-image
make run-image
[root@c52bfa555d50 ~]# bash buildJDK.sh
[root@c52bfa555d50 ~]# /opt/mandrelJDK/bin/native-image --version
GraalVM Version dev (Java Version 11.0.8-ea)
```

galderz:
@zakkak That looks excellent! Let me give it one last try :)

galderz:
Nice work @zakkak!

PR #17, opened 2020-05-29
Add script to build mandrelJDK

### LINKED ISSUE #16: Add functionality to produce a mandrelJDK
Currently mandrel-packaging only builds the maven artifacts.
It should be extended to build a full mandrelJDK (JVM + native bits), possibly generating a zip file.

zakkak:
This is more or less a copy of https://github.com/zakkak/kolitiri/blob/master/build.sh which I have been using to build a mandrelJDK.

I have edited it to work with the container images created by this repository as well as locally.

e.g. to build a mandrelJDK locally you can use:
`MANDREL_REPO=~/code/graal JAVA_HOME=/opt/jvms/openjdk-11.0.8+4/ MANDREL_JDK=./mandrelJDK MX_HOME=~/code/mx ./src/buildJDK.sh`

to do the same in the container just run `./src/build.sh`

Closes #16

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Building mandrelJDK locally
 
 ```shell
-JAVA_HOME=/opt/jvms/openjdk-11.0.8+4/ MX_HOME=~/code/mx MANDREL_REPO=~/code/mandrel  MANDREL_JDK=./mandrelJDK  ./src/buildJDK.sh
+JAVA_HOME=/opt/jvms/openjdk-11.0.8+4/ MX_HOME=~/code/mx MANDREL_REPO=~/code/mandrel  MANDREL_JDK=./mandrelJDK  ./buildJDK.sh
 ```
 
 where:
--- a/README.md
+++ b/README.md
@@ -1,14 +1,14 @@
 # Building mandrelJDK locally
 
 ```shell
-JAVA_HOME=/opt/jvms/openjdk-11.0.8+4/ MX_HOME=~/code/mx MANDREL_REPO=~/code/mandrel  MANDREL_JDK=./mandrelJDK  ./buildJDK.sh
+JAVA_HOME=/opt/jvms/openjdk-11.0.8+4/ MX_HOME=~/code/mx MANDREL_REPO=~/code/mandrel MANDREL_HOME=./mandrelJDK  ./buildJDK.sh
 ```
 
 where:
 * `JAVA_HOME` is the path to the OpenJDK you want to use for building mandrel
 * `MX_HOME` is the path where you cloned https://github.com/graalvm/mx
 * `MANDREL_REPO` is the path where you cloned https://github.com/graalvm/mandrel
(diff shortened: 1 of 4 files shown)


zakkak:
One things that needs to be improved is that It doesn't use `src/build.java` to build the JVM bits.
I suggest making `src/build.java` get the paths from env variables to make it work without the need of a container image.

galderz:
I agree on extending `build.java` to help with native image building, but why from env variables? Why not command line parameters? 

`build.java` supports command line parameters (see [here](https://github.com/mandrel/mandrel-packaging/blob/master/run-image.sh#L37-L52) for the options that are often passed in), it can be easily extended to add more command line options.

Command line args are easier to unit test as well.

zakkak:
:+1: I guess went for env variables because that's what I am using in the shell script, we can use parameters in the shell script as well though :)

buildJDK.sh
```diff
@@ -0,0 +1,71 @@
...
+
+MX_HOME=${MX_HOME:-/opt/mx}
+JAVA_HOME=${JAVA_HOME:-/opt/jdk}
+MANDREL_REPO=${MANDREL_REPO:-/tmp/mandrel}
+MANDREL_JDK=${MANDREL_JDK:-/opt/mandrelJDK}
+
+### Build Mandrel
+## JVM bits
+basename="$(dirname $0)"
+${JAVA_HOME}/bin/java -ea $basename/src/build.java --version 20.1.0.redhat-00001 --maven-local-repository /tmp/.m2/repository --mx-home ${MX_HOME} --mandrel-home ${MANDREL_REPO}
```
galderz, line 15: You could pass `${VERBOSE}` to `build.java` so that you can also see verbose output there.

README.md
```diff
@@ -1,3 +1,20 @@
...
+
+```shell
+JAVA_HOME=/opt/jvms/openjdk-11.0.8+4/ MX_HOME=~/code/mx MANDREL_REPO=~/code/mandrel  MANDREL_JDK=./mandrelJDK  ./buildJDK.sh
+```
+
+where:
+* `JAVA_HOME` is the path to the OpenJDK you want to use for building mandrel
+* `MX_HOME` is the path where you cloned https://github.com/graalvm/mx
+* `MANDREL_REPO` is the path where you cloned https://github.com/graalvm/mandrel
+* `MANDREL_JDK` is the path where you want mandrel to be installed, after completion you will be
```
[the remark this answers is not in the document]
zakkak: fixed in https://github.com/mandrel/mandrel-packaging/pull/17/commits/fab25eb8501a12d7cdfe4ac922b8a0860ddf172a

buildJDK.sh
zakkak: Fixed in https://github.com/mandrel/mandrel-packaging/pull/17/commits/cf6f5a1f72acf5fa57a87d658f9d3cc308bdb72f

Karm:
@galderz Could you merge it, pls?

Issue #19, opened 2020-06-05
Set up a github workflow to test mandrel-packaging with different options/scenarios


zakkak:

zakkak:
We now have some basic coverage for the build and ansible scripts.

PR #26, opened 2020-06-18
Add github action for testing

zakkak:
Implements part of #19

### DIFF
--- a/.github/workflows/buildJDK.yml
+++ b/.github/workflows/buildJDK.yml
@@ -1,0 +1,80 @@
+name: Build Mandrel using buildJDK and test it
+
+on: [push]
+
+env:
+  MX_GIT_CACHE: refcache
+  JAVA_HOME: /opt/jdk
+  MANDREL_REPO: ${{ github.workspace }}/mandrel
+  MX_HOME: ${{ github.workspace }}/mx
+  VERBOSE: true
+  MANDREL_VERSION: 10.9.8.7 test-build
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/buildJDK.yml
```diff
@@ -0,0 +1,78 @@
...
+                );
+            }
+        }
+        " > HelloStrict.java
+        /opt/mandrelJDK/bin/javac HelloStrict.java
+        /opt/mandrelJDK/bin/java HelloStrict > java.txt
+        /opt/mandrelJDK/bin/native-image HelloStrict
+        ./hellostrict > native.txt
+        diff java.txt native.txt
+    - name: Upload Mandrel build
```
jerboaa, line 74: Do we need to upload this test build? If so why? My concern is that those test version bits `10.9.8.7` could be confused for real versions. I'd rather avoid uploading it altogether. The smoke tests are good enough for me for now.

zakkak: This is an artifact upload for the CI (this will not be visible at releases or such). 
To get access to this tarball you have to open the CI job go to artifacts and download it (something kind of hard to do by mistake).
It's only useful to inspect the created tarball if something seems odd, or even run some extra tests locally for debugging.

The random version is just to check that versioning works OK.

I don't mind removing the upload part or changing any other part if you still find this confusing.
Note that we do the same in https://github.com/graalvm/mandrel/blob/master/.github/workflows/quarkus.yml#L53-L57 and there we cannot avoid it because we need the artifact in a later stage of the CI.

jerboaa: OK. In that case I suggest to upload it without a version like for https://github.com/graalvm/mandrel/blob/master/.github/workflows/quarkus.yml#L53-L57. At least I'd prefer that.

PR #33, opened 2020-06-22
Ansible updates plus GH workflow and a some small enhancements to buildJDK.sh

zakkak:
* Updates Ansible to work with latest openJDK and mandrel's new branching scheme.
* Adds GH workflow for testing docker image created with ansible
* Small enhancements to buildJDK.sh


zakkak:
Closes #19

PR #46, opened 2020-06-29
Tests buildJDK on both master and mandrel/20.1

zakkak:
Since master is expected to fail now and then (see https://github.com/oracle/graal/issues/2628) we should test against both master and the latest stable branch.

### DIFF
--- a/.github/workflows/buildJDK.yml
+++ b/.github/workflows/buildJDK.yml
@@ -6,12 +6,14 @@ on:
       - 'buildJDK.sh'
       - 'src/**'
       - 'resources/**'
+      - '.github/workflows/buildJDK.yml'
   pull_request:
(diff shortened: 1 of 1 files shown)


galderz:
Looks good @zakkak but CI failed. I couldn't easily see what's wrong though :\

zakkak:
The CI fails on master are caused by https://github.com/oracle/graal/issues/2628
The backport is on it's way :)

galderz:
@zakkak Ah ok, wasn't clear from the outputs of CI. Do we wait until that's in place to integrate this?

zakkak:
I say no, and that's the motivation behind this PR. We need to be able to test mandrel-packaging even if master is broken (due to some unrelated to mandrel-packaging issue).

PR #51, opened 2020-07-15
Fix ansible playbook

zakkak:
fedora:32 no longer provides the python symlink so we need to
explicitly set MX_PYTHON so that mx will use python3 instead of python

### DIFF
--- a/ansible/roles/mx/tasks/main.yml
+++ b/ansible/roles/mx/tasks/main.yml
@@ -32,7 +32,9 @@
     extra_opts:
     - --strip-components=1
 
-- name: Add mx to $PATH
+- name: Add mx to $PATH and set MX_PYTHON
   copy:
     dest: /etc/profile.d/mx.sh
-    content: "export PATH={{ mx_path }}:$PATH"
+    content: |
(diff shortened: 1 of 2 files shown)


ansible/roles/mx/tasks/main.yml
```diff
@@ -32,7 +32,7 @@
     extra_opts:
     - --strip-components=1
 
-- name: Add mx to $PATH
+- name: Add mx to $PATH and set MX_PYTHON
   copy:
     dest: /etc/profile.d/mx.sh
-    content: "export PATH={{ mx_path }}:$PATH"
+    content: "export PATH={{ mx_path }}:$PATH\nexport MX_PYTHON=python3"
```
Karm, line 38: @zakkak  You sure about the \n ?
Isn't it 
```
     dest: /etc/profile.d/mx.sh 
     content: |
        line one
        line two
```
?  Not sure. Just took a peak into some Ansible yaml tutorial. I am not a user myself.

zakkak: Yes this is more of a yaml think, not ansible, but you are right it will be more clear that way.

PR #56, opened 2020-08-25
Add options to skip more steps when packaging

zakkak:
Namely, archiving and `maven install`


jerboaa (approved):
Yes, please!

galderz:
I was just thinking that yesterday. Good idea!

zakkak:
In just noticed however that this will also not update the jar files used to create the JDK, since `buildJDK.sh` uses the jars from the maven repo and not the local directories. This is fixed in #57 however, so I don't think it's worth fixing in `buildJDK.sh`

Issue #55, opened 2020-08-20
Support building Mandrel on aarch64


galderz:
Fails with:

```
cp: cannot stat '/home/g/.qollider/cache/2008/mandrel/substratevm/mxbuild/linux-amd64/src/com.oracle.svm.native.libchelper/amd64/liblibchelper.a': No such file or directory
Exception in thread "main" java.lang.RuntimeException: java.lang.RuntimeException: Failed, exit code: 1
	at OperatingSystem.exec(qollider.java:1532)
	at Steps$Exec.run(qollider.java:1197)
	at Mandrel.build(qollider.java:691)
	at SystemMandrel.build(qollider.java:641)
	at qollider.main(qollider.java:91)
Caused by: java.lang.RuntimeException: Failed, exit code: 1
	at OperatingSystem.exec(qollider.java:1526)
	... 4 more
```

`buildJDK.sh` is assuming `amd64`, see [here](https://github.com/graalvm/mandrel-packaging/blob/master/buildJDK.sh#L95).

Karm:
The build works: https://karms.biz/tmp/mandrel-java11-linux-aarch64-20.2-SNAPSHOT.tarxz
What we need now is a CI setup. Then this issue can be closed.

Karm:
This issue is alive and kicking :)

Issue #54, opened 2020-08-11
Port the build harness to Windows


Karm:
The result should be a native-image capable JDK, built using pure MSVC toolchain (no cygwin et al.)

Karm:
Very much alive. Removed Stale label.

Karm:
Resolved by #71

PR #57, opened 2020-08-25
Move to java

### LINKED ISSUE #48: "git -C" not always present
On older systems, like Centos 7, `git -C` is not an option:

```
Unknown option: -C
usage: git [--version] [--help] [-c name=value]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p|--paginate|--no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           <command> [<args>]
```

This is used to compute the mandrel version [here](https://github.com/graalvm/mandrel-packaging/blob/master/buildJDK.sh#L16).

zakkak:
Closes #48
Moves towards #54 and #55 

Please note that this probably needs some more love (refactoring), so please keep the comments coming :)

Also note that this will disrupt your workflow once merged since `buildJDK.sh` will no longer be a thing!

A few highlights:
1. maven install is no longer run by default 
2. Archiving is no longer run by default 
3. Most parts have been implemented in Java except for archiving and getting the version through `git`
4. We no longer create tarxz files but tar.xz (@Karm is this OK?)


.github/workflows/buildJDK.yml
```diff
@@ -20,17 +16,17 @@ env:
...
   build-and-test:
-    name: Build and test ${{ matrix.mandrel-ref }} branch/tag
-    runs-on: ubuntu-latest
+    name: Build and test ${{ matrix.mandrel-ref }} on ${{ matrix.os}}
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        mandrel-ref: [graal/master, mandrel/20.1]
+        os: ['ubuntu-latest', 'macos-latest']
```
jerboaa, line 28: This won't work unless you find an OpenJDK provider with static libs built for MacOS X. The current setup also hard-codes the linux OS in the AdoptOpenJDK API call.

zakkak: Correct, I was just trying to test if the packaging works on macOS as well. 
I am going to remove this commit and open an issue in graalvm/mandrel to start a discussion around what's needed to get macOS support.

zakkak: https://github.com/graalvm/mandrel/issues/114

build.java
```diff
@@ -28,18 +22,135 @@
...
         final var options = Options.from(Args.read(args));
 
         final var fs = FileSystem.ofSystem(options);
         final var os = new OperatingSystem();
         final var build = new SequentialBuild(fs, os);
+        final var mandrelRepo = FileSystem.mandrelRepo(options);
+
+        if (options.mandrelVersion == null) {
+            // git -C ${MANDREL_REPO} describe 2>/dev/null || git -C ${MANDREL_REPO} rev-parse --short HEAD) | sed 's/mandrel-//'
+            ProcessBuilder builder = new ProcessBuilder();
```
galderz, line 37: I wonder whether these invocations could be hidden behind an overloaded `OperatingSystem.exec` method that returns the String output of the command executed.

Also, I see the potential opportunity for `ProcessBuilder` to be a final instance var of `OperatingSystem` and the have different `exec` invocations work on that.

build.java
```diff
@@ -1,23 +1,17 @@
...
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Queue;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.*;
```
galderz, line 6: Not sure we want to bring the kitchen sink.

build.java
```diff
@@ -1,23 +1,17 @@
...
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.*;
```
galderz, line 8: Neither here.

galderz (commented):
I'm pleased to see that we consolidate the build in the Java class.

I think the code is fine as is, but I feel there's room for improvement here.

Aside from the comments on the code, I wanted to stress that the way `FileSystem`, `OperatingSystem` and `Tasks.*` work, they've been designed to separate the inputs from the external dependencies or effects (working on the FS, invoking an exec on the OS...etc) and enable easy unit testing as demonstrated in the `Check` class. Ideally, I'd like to see any new added functionality enhance the `Check` unit tests to make sure we don't break things in the future and enable quicker feedback on changes here.

build.java
zakkak: Good point. Please see new version :)

jerboaa (approved):
OK with me.

galderz (approved):
Thx for the improvements @zakkak :)

Karm:
@zakkak  Could you include a "migration guide" comment here, please?

```
./buildJDK.sh: No such file or directory
```
  
For one who previously used to do:
```
export MAVEN_REPO=${HOME}/.m2/repository
export MANDREL_REPO=${WORKSPACE}/mandrel
export JAVA_HOME=/usr/java/${OPENJDK}
export MX_HOME=${WORKSPACE}/mx
export M2_HOME=/opt/apache-maven-3.6.3
export PATH=${JAVA_HOME}/bin:${M2_HOME}/bin:${MX_HOME}:${PATH}
pushd mandrel
export MANDREL_VERSION="${MANDREL_VERSION_SUBSTRING} $( git log --pretty=format:%h -n1)"
export MANDREL_VERSION_UNTIL_SPACE="$( echo ${MANDREL_VERSION} | sed -e 's/\([^ ]*\).*/\1/;t' )"
export JAVA_VERSION="$(java --version | sed -e 's/.*build \([^) ]*\)).*/\1/;t;d' )"
export MANDREL_HOME=${WORKSPACE}/mandrel-java11-${MANDREL_VERSION_UNTIL_SPACE}
popd
./buildJDK.sh
TAR_NAME="$( ls mandrel-*.tar.gz )"
sha1sum ${TAR_NAME}>${TAR_NAME}.sha1
cat >./MANDREL.md <<EOL
This is a dev build of Mandrel from https://github.com/graalvm/mandrel.
Mandrel ${MANDREL_VERSION}
OpenJDK used: ${JAVA_VERSION}
EOL
if [[ ! -e "${MANDREL_HOME}/bin/native-image" ]]; then
  echo "Cannot find native-image tool. Quitting..."
  exit 1
fi
cat >./Hello.java <<EOL
public class Hello {
public static void main(String[] args) {
    System.out.println("Hello.");
}
}
EOL
export JAVA_HOME=${MANDREL_HOME}
export PATH=${JAVA_HOME}/bin:${PATH}
javac Hello.java
native-image Hello
if [[ "`./hello`" == "Hello." ]]; then echo Done; else echo Native image fail;exit 1;fi
```

zakkak:
@Karm in the above example you can replace `./buildJDK.sh` with
```
${JAVA_HOME}/bin/java -ea build.java --maven-local-repository ${MAVEN_REPO} --mandrel-repo ${MANDREL_REPO} --mx-home ${MX_HOME} --maven-home ${M2_HOME} --mandrel-version ${MANDREL_VERSION} --mandrel-home ${MANDREL_HOME} --archive-suffix tar.gz
```

Karm:
@zakkak  Ad 

 > We no longer create tarxz files but tar.xz (@Karm is this OK?)

That is not O.K. :(, the internal infrastructure has been wired around `.tarxz` suffix for years, it parses for it on numerous places etc. If we change it here, I would have to go and revert/patch it internally to do `.tarxz`...which is kinda pointless given the fact we added .tarxz for the internal infra in the first place.

Karm (approved):
LGTM!   Thx.
Build works with the change: https://ci.modcluster.io/job/mandrel-20.2-linux-build-packaging-moved-to-java/1/

PR #64, opened 2020-10-08
Fix race that sometimes results in missing git output

zakkak:
Getting the input stream of the process after starting it allows the
process to output data before we start capturing them resulting in cases
where we miss the output of git-describe etc. With this patch we define
where we want to capture the output before starting the process, thus
avoiding the above issue.


Karm (approved):
:+1:

PR #71, opened 2020-11-03
Adds support for building Mandrel on Windows

Karm:
- Windows support
- zip archive support for both Linux and Windows
- formatting according to the committed code-formatting-rules.xml
- some type/stylistic changes that made it easier for me to work with it but I can revert...

Building on Windows
------------------

* Installed Visual Studio (C/C++ support)
* Installed Python 2.7 and on PATH  (de facto mx requirements)
* Downloaded OpenJDK
* git et co., e.g. from https://github.com/cmderdev/cmder/releases/download/v1.3.16/cmder.zip
* Have on PATH: C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build
Then:
```
set PYTHONHTTPSVERIFY=0
vcvarsall.bat x86_amd64
set JAVA_HOME=C:\projects\openjdk-11.0.9_11\
set PATH=%JAVA_HOME%\bin;%PATH%
set MAVEN_REPO=C:/projects/.m2
set MANDREL_REPO=C:/projects/graal/graal
set MX_HOME=C:/projects/graal/mx
set MANDREL_HOME=C:/projects/graal/mandrel-java11-windows-amd64-20.2-SNAPSHOT
%JAVA_HOME%/bin/java -ea build.java --maven-local-repository %MAVEN_REPO% --mandrel-repo %MANDREL_REPO% --mx-home %MX_HOME% --mandrel-version 20.2-SNAPSHOT --mandrel-home %MANDREL_HOME% --archive-suffix zip --verbose
```

# Status

It works, it builds Quarkus app on Windows and Linux (i.e. Linux build is not broken) in a "clean room" env. I will be doing some additional work though:
 * fix asserts in ```     static void checkMx() ``` method
 * cleanup some TODOs, e.g. superfluous flag ```--force-bash-launchers=false```
 * running the Mandrel build in clean room to make absolutely certain that the dependencies and prerequisites are only those listed above

The overall picture is working and O.K. though....

### DIFF
--- a/build.java
+++ b/build.java
@@ -3,23 +3,26 @@ import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
 import java.nio.file.FileVisitOption;
 import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
+import java.nio.file.StandardOpenOption;
 import java.nio.file.attribute.BasicFileAttributes;
-import java.util.Arrays;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Queue;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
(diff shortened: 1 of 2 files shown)


build.java
```diff
@@ -36,141 +38,246 @@
...
+        }
+        final Path nativeImage = mandrelHome.resolve(Path.of("lib", "svm", "bin", IS_WINDOWS ? "native-image.cmd" : "native-image"));
+
+        logger.debugf("Copy native-image...");
+        FileSystem.copy(mandrelRepo.resolve(
+            Path.of("sdk", "mxbuild", PLATFORM, IS_WINDOWS ? "native-image.exe.image-bash" : "native-image.image-bash", IS_WINDOWS ? "native-image.cmd" : "native-image")),
+            nativeImage);
+        if (IS_WINDOWS)
+        {
+            FileSystem.copy(mandrelRepo.resolve(
```
zakkak, line 124: That's the second time this gets copied :)

build.java
```diff
@@ -36,141 +38,246 @@
...
+                fs.workingDir());
+        }
+        else if (archiveName.endsWith("tarxz"))
+        {
+            archiveCommand = Tasks.Exec.of(
+                Arrays.asList("tar", "cJf", archiveName, "-C", mandrelHomeParentDir, mandrelHomeDirName),
+                fs.workingDir(),
+                new EnvVar("XZ_OPT", "-9e"));
+        }
+        else if (archiveName.endsWith("zip") && IS_WINDOWS)
```
zakkak, line 177: using https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/zip/ZipOutputStream.html instead would make this platform agnostic and also remove any external dependencies

build.java
```diff
@@ -36,141 +38,246 @@
...
+        boolean foundWindowsCPPattern = false;
+        for (int i = 0; i < lines.size(); i++)
+        {
+            if (IS_WINDOWS && !foundWindowsCPPattern)
+            {
+                final Matcher windowsCPMatcher = windowsCPPattern.matcher(lines.get(i));
+                if (windowsCPMatcher.find())
+                {
+                    final String relcp = lines.get(i)
+                        .replace("..\\..\\graalvm\\", "..\\lib\\graalvm\\")
```
zakkak, line 234: You shouldn't need to do this.
To avoid this you should copy the native-image.cmd in `lib\svm\bin` and then create a symlink in `bin\`

zakkak: This will simplify the code in https://github.com/graalvm/mandrel-packaging/pull/71/files#diff-9d0a8c0df070b9b21cdbb6b730b6b8510955741116b6e63cdab769918899d203R242-R280 as well

build.java
```diff
@@ -518,7 +634,11 @@ public static Effects ofSystem()
 
             public static Effects noop()
             {
-                return new Effects(p -> Stream.empty(), (p, l) -> {}, (src, target) -> {});
+                return new Effects(p -> Stream.empty(), (p, l) ->
```
zakkak, line 637: well that's ugly :D

build.java
```diff
@@ -536,12 +656,23 @@ public static Effects noop()
...
-            BuildArgs.of("--projects", "com.oracle.svm.native.libchelper,com.oracle.svm.native.jvm.posix")
-            , BuildArgs.of("--only", "native-image.image-bash,native-image-agent-library_native-image.properties")
+        BuildArgs.of("--projects",
+            build.IS_WINDOWS ?
+                "com.oracle.svm.native.libchelper,com.oracle.svm.native.jvm.windows,org.graalvm.nativeimage,com.oracle.svm.core.windows" :
+                "com.oracle.svm.native.libchelper,com.oracle.svm.native.jvm.posix")
+        , BuildArgs.of("--only",
+            build.IS_WINDOWS ?
+                // "native-image.exe.image-bash,native-image-configure.exe.image-bash,native-image-agent-library_native-image.properties" :
+                "native-image.exe.image-bash,native-image.exe.image-bash-stage1,native-image-configure.exe.image-bash,native-image-configure.exe.image-bash-stage1,native-image-agent-library_native-image.properties,native-image-configure-launcher_native-image.properties,native-image-launcher_native-image.properties" :
```
zakkak, line 674: Do we really need to build all these?

build.java
```diff
@@ -600,15 +733,16 @@ static void build(
...
+            final Path mx = mxHome.apply(Paths.get("mx"));
+            final List<String> args = Lists.concat(
                 List.of(
                     mx.toString()
                     , options.verbose ? "-V" : ""
                     , "--trust-http"
                     , "--java-home"
                     , javaHome.get().toString()
                     , "--native-images=lib:native-image-agent"
+                    , "--force-bash-launchers=false" // TODO, Fishy?
```
zakkak, line 745: Probably not needed, at least in linux by default native-image launchers are not native.

build.java
```diff
@@ -677,15 +818,12 @@ static void swapDependencies(Options options, Tasks.FileReplace.Effects effects,
     private static Function<Stream<String>, List<String>> swapDependencies(List<Dependency> dependencies)
     {
         return lines ->
-        {
-            ParsedDependencies parsed = parseSuitePy(dependencies, lines);
-            return applyDependencies(dependencies, parsed);
-        };
+            applyDependencies(dependencies, parseSuitePy(dependencies, lines));
```
zakkak, line 821: nitpick: I would merge this with the previous line

build.java
```diff
@@ -807,19 +945,19 @@ private static boolean notMavenOrg(String line)
     private static Function<String, String> prependMavenProxy(Options options)
     {
         return line ->
-        {
-            var mavenBaseURL = String.format("\"%s/\"", options.mavenProxy);
-            return line.contains("_mavenRepoBaseURLs")
-                ? line.replaceFirst("\\[", String.format("[ %s", mavenBaseURL))
+            line.contains("_mavenRepoBaseURLs")
+                ? line.replaceFirst("\\[", String.format("[ %s", String.format("\"%s/\"", options.mavenProxy)))
```
zakkak, line 949:
```suggestion
                ? line.replaceFirst("\\[", String.format("[ \"%s/\"", options.mavenProxy))
```
Isn't this essentially the same?

build.java
```diff
@@ -1585,18 +1746,19 @@ public static void main(String... args)
...
+        final Tasks.FileReplace.Effects replace = Tasks.FileReplace.Effects.noop();
         final Function<Path, Path> identity = Function.identity();
         final Supplier<Path> javaHome = () -> Path.of("java");
         Mx.build(options, exec, replace, identity, identity, javaHome);
         os.assertNumberOfTasks(4);
         os.assertTask("mx clean");
-        os.assertTask("mx --trust-http --java-home java --native-images=lib:native-image-agent --exclude-components=nju build --no-native --dependencies SVM,SVM_DRIVER,SVM_AGENT");
-        os.assertTask("mx --trust-http --java-home java --native-images=lib:native-image-agent --exclude-components=nju build --projects com.oracle.svm.native.libchelper,com.oracle.svm.native.jvm.posix");
-        os.assertTask("mx --trust-http --java-home java --native-images=lib:native-image-agent --exclude-components=nju build --only native-image.image-bash,native-image-agent-library_native-image.properties");
+        // TODO fix assert
```
zakkak, line 1758: I am personally OK with removing those tests. Ideally we would need some form of JUnit tests.

zakkak (changes requested):

zakkak:
Nice work @Karm :+1:

build.java
```diff
@@ -36,141 +38,246 @@
...
+        if (IS_WINDOWS)
+        {
+            FileSystem.copy(mandrelRepo.resolve(
+                Path.of("sdk", "mxbuild", PLATFORM, "native-image.exe.image-bash", "native-image.cmd")),
+                mandrelHome.resolve(Path.of("lib", "svm", "bin", "native-image.cmd")));
+        }
 
-        // logger.info("Copy macros!");
+        logger.debugf("Symlink native-image...");
+        // Windows: Seriously? I had to run cmd as Administrator to execute this without getting "A required privilege is not held by the client."
```
zakkak, line 130: The symlinking of native-image to native-image.cmd seems redundant. I downloaded graalvm CE for windows and installed native image with `gu` and it only includes `lib\svm\bin\native-image.exe` and `bin\native-image.cmd` (no `native-image`).

Both `native-image --version` and `native-image.cmd --version` work fine.

build.java
jerboaa: Isn't the intention the opposite? I.e. get a `native-image.exe` instead of a cmd file?

zakkak: Not necessarily and judging by the changes it looks like @Karm managed to make it work with the cmd script.
Note that in Linux we currently ship a script so shipping a binary for windows might introduce one more dimension in our matrix of differences. I would prefer to either ship both with either a binary native-image or a script native-image.

build.java
```diff
@@ -36,141 +38,246 @@
...
 
-    public static void main(String... args) throws IOException {
+    public static void main(String... args) throws IOException
+    {
         Check.main();
 
-        final var options = Options.from(Args.read(args));
+        final Options options = Options.from(Args.read(args));
 
-        final var fs = FileSystem.ofSystem(options);
```
galderz, line 45: I see no reason to change this :\

galderz (changes requested):

build.java
Karm: THX. Removing.

build.java
Karm: This is intentional and I would very much like it to stay that way. I had been having a lot of problems with JDK zip over e.g. NFS drives on Windows or on Windows VMs with very slow I/O and I resulted to using it only if I have the data in JVM memory already, i.e. if I am streaming to disk something I created in the program.  To use native OS zip utils to zip a directory on the disk has proven to me to be the most stable approach.  The only "external dependency" is zip on Linux which might not be a default part of your distro.  Powershell zip capability is present by default on all Windows systems we are targeting.

build.java
Karm: @zakkak  You are not correct. It is more complicated than that. It caused me a lot of trouble to realize what was going on. I will try to explain the problem and perhaps if you or @jerboaa have more powerful Windows-fu, we might find a way to simplify it.

# Linux
Consider this script:

```bash
karm@localhost:/tmp/X$ cat Y/Z/script.sh
#!/bin/sh
source="${BASH_SOURCE[0]}"
while [ -h "$source" ] ; do
    prev_source="$source"
    source="$(readlink "$source")";
    if [[ "$source" != /* ]]; then
        # if the link was relative, it was relative to where it came from
        dir="$( cd -P "$( dirname "$prev_source" )" && pwd )"
        source="$dir/$source"
    fi
done
location="$( cd -P "$( dirname "$source" )" && pwd )"

echo My location is ${location}.
```

Symlinked thus:
```
karm@localhost:/tmp/X$ ls -lah
total 12K
drwxrwxr-x.  3 karm karm   32 Nov  4 11:37 .
drwxrwxrwt. 44 root root 8.0K Nov  4 12:06 ..
lrwxrwxrwx.  1 karm karm   13 Nov  4 11:37 script.sh -> Y/Z/script.sh
drwxrwxr-x.  3 karm karm   15 Nov  4 11:35 Y
```

When you call it, you get the *location of the target*:

```
karm@localhost:/tmp/X$ ./script.sh
My location is /tmp/X/Y/Z.
```

# Windows

Consider this script:

```cmd
@echo off
setlocal enabledelayedexpansion
call :getScriptLocation location

echo My location is %location%.

goto :eof

:getScriptLocation variableName
    set "%~1=%~dp0"
    exit /b 0

```

Symlinked thus:

```
C:\tmp\X
λ dir
 Volume in drive C has no label.
 Volume Serial Number is E0DC-908D

 Directory of C:\tmp\X

11/03/2020  08:05 AM    <DIR>          .
11/03/2020  08:05 AM    <DIR>          ..
11/03/2020  08:05 AM    <SYMLINK>      script.cmd [Y\Z\script.cmd]
11/03/2020  07:42 AM    <DIR>          Y
               1 File(s)              0 bytes
               3 Dir(s)  31,521,820,672 bytes free

```

When I call the script, I get the *location of the symlink, not the target*:

```
C:\tmp\X
λ script
My location is C:\tmp\X\.
```

So, what you can see as somewhat desperate replacing the paths is merely a solution to this discrepancy in behaviour. To have the script actually reside in `lib/svm/bin` and to call it from `bin`, one has to consider the paths originating from the `bin` on Windows...

I am open for suggestions, but I cannot just remove it as t is now.

build.java
Karm: Sorry...gonna prettify it :-)

build.java
Karm: Thanks for pointing it out. I will clean it up....

build.java
Karm: Will remove it and test again.

build.java
Karm: It is.   Formatting...  Will revert if you find it better.

build.java
Karm: I don't think it makes sense to have anything like that here. The test for the script is doing the actual build and running it with --verbose and parsing the log for unexpected warnings and unexpected errors. I think the mandrel-intergtaion-tests harness is a good candidate for it.

What should e present in the script is some basic environment check: Do we have cl.exe? Do we have python? Mx on path? etc. So as it fails soon with some hint.

build.java
Karm: When it comes to symlinking, it is problematic, see: https://github.com/graalvm/mandrel-packaging/pull/71#discussion_r517268470

We don't have native-image.exe. The build script as it is written now does not produce it.

build.java
Karm: It is a matter of style. It makes it subjectively unnecessarily more difficult to me to navigate the script with `var` all over the place.   If you feel strongly about having it there, I am O.K. with reverting those lines. The point of this PR is Windows support and this change is obviously not essential for that.

build.java
zakkak: I see! I was not aware of that. Thanks for the explanation. 

I see that GraalVM CE instead of symlinking creates a proxy script that just calls `lib\svm\bin\native-image.exe`, I guess we can do the same for `lib\svm\bin\native-image.cmd`

The proxy script looks like this:
```
@echo off

call :getScriptLocation location
"%location%..\lib\svm\bin\native-image.exe" %*
goto :eof

:: If this script is in `%PATH%` and called quoted without a full path (e.g., `"js"`), `%~dp0` is expanded to `cwd`
:: rather than the path to the script.
:: This does not happen if `%~dp0` is accessed in a subroutine.
:getScriptLocation variableName
    set "%~1=%~dp0"
    exit /b 0
```

build.java
zakkak: I think it's "simpler" and more "readable". That's my opinion though... feel free to keep either version :)

build.java
zakkak: Since this "script" is now almost 2k LOCs some unit tests would become handy when tweaking it, but sure the runtime checks for dependencies would be much more useful :)

build.java
zakkak: Rechecking my comment here and it doesn't make any sense, I guess I missread the code :/

build.java
Karm: @zakkak The proxy script sounds like the best (simplest) way. Lemme try it...

build.java
zakkak: OK, I didn't know about such issues. Additionally, since zip is expected to be used only on Windows and it's available by default, it sounds fine to me :)

build.java
zakkak: I just noticed that we might not be talking about the same thing here.
When I asked if this is essentially the same I was talking about my suggestion in https://github.com/graalvm/mandrel-packaging/pull/71#discussion_r516879370 which inlines `String.format("\"%s/\"", options.mavenProxy)`.

Your initial change that inlines `mavenBaseURL` is totally fine!

build.java
zakkak: Since this makes @Karm's life easier and doesn't seem to impact the rest of the team that much we decided that it's OK to avoid using `var` in `build.java`.

Karm:
@zakkak @galderz @jerboaa Ready for review. I think it is ready to be merged.
I tested with Mandrel 20.2 on Windows 10 and Centos 8.

build.java
```diff
@@ -36,141 +39,239 @@
...
+        final Path nativeImage = mandrelHome.resolve(Path.of("lib", "svm", "bin", IS_WINDOWS ? "native-image.cmd" : "native-image"));
+
+        logger.debugf("Copy native-image...");
+        FileSystem.copy(mandrelRepo.resolve(
+            Path.of("sdk", "mxbuild", PLATFORM, IS_WINDOWS ? "native-image.exe.image-bash" : "native-image.image-bash", IS_WINDOWS ? "native-image.cmd" : "native-image")),
+            nativeImage);
+        // We don't create symlink on Windows, See https://github.com/graalvm/mandrel-packaging/pull/71#discussion_r517268470
+        if (IS_WINDOWS)
+        {
+            logger.debugf("Generate native-image.cmd...");
```
zakkak, line 126: You can (and probably should) get the template from `sdk/mx.sdk/vm/exe_link_template.cmd`  and only replace `<target>`.

build.java
```diff
@@ -36,141 +39,239 @@
...
-        // logger.info("Copy macros!");
+        logger.debugf("Copy macros...");
         fs.copyDirectory(mandrelRepo.resolve(Path.of("sdk", "mxbuild", "native-image.properties", "native-image-agent-library")),
-                mandrelHome.resolve(Path.of("lib", "svm", "macros", "native-image-agent-library")));
+            mandrelHome.resolve(Path.of("lib", "svm", "macros", "native-image-agent-library")));
 
-        // Patch native image
-        patchNativeImageLauncher(nativeImage, options.mandrelVersion);
+        logger.debugf("Patch native image...");
+        if (IS_WINDOWS)
```
zakkak, line 153: If I am not mistaken this if else is no longer needed since nativeImage is properly set for both OSes in https://github.com/graalvm/mandrel-packaging/pull/71/files#diff-9d0a8c0df070b9b21cdbb6b730b6b8510955741116b6e63cdab769918899d203R117

build.java
```diff
@@ -36,141 +39,239 @@
...
-                newLine.append(" --add-modules \"org.graalvm.truffle,org.graalvm.sdk\"");
-                newLine.append(" --module-path ${location}/../../truffle/truffle-api.jar:${location}/../../jvmci/graal-sdk.jar");
-                newLine.append(matcher.group(2));
-                newLine.append(" -J--add-exports=jdk.internal.vm.ci/jdk.vm.ci.code=jdk.internal.vm.compiler");
-                newLine.append(matcher.group(3));
-                lines.set(i, newLine.toString());
+    private static void patchNativeImageLauncher(Path nativeImage, String mandrelVersion) throws IOException
+    {
+        final List<String> lines = Files.readAllLines(nativeImage);
+        // This is jamming two sets of parameters in between three sections of command line.
```
zakkak, line 236: I am trying to fix this upstream, see https://github.com/oracle/graal/pull/2962
For the time being I prefer it to be fragile and crash if anything changes rather than generating our own line and letting some change go unnoticed.

build.java
```diff
@@ -536,12 +648,25 @@ public static Effects noop()
...
     );
 
     static final List<BuildArgs> BUILD_NATIVE_STEPS = List.of(
-            BuildArgs.of("--projects", "com.oracle.svm.native.libchelper,com.oracle.svm.native.jvm.posix")
-            , BuildArgs.of("--only", "native-image.image-bash,native-image-agent-library_native-image.properties")
+        BuildArgs.of("--projects",
+            build.IS_WINDOWS ?
+                "com.oracle.svm.native.libchelper," +
+                    "com.oracle.svm.native.jvm.windows," +
+                    "org.graalvm.nativeimage," +
```
zakkak, line 659: The `org.graalvm.nativeimage` seems non-windows specific. It's strange that it's required only on Windows :/
Do you recall why it's needed?

zakkak (changes requested):
Thanks for the changes @Karm please see a couple more comments.

build.java
Karm: Ah! Right.  Will do.

Karm:
> Thanks for the changes @Karm please see a couple more comments.

@zakkak I addressed the comments. Thanks for pointing out the template, it didn't occur tome it's there...      
I also removed the superfluous if-else branched and the redundant build module.

zakkak (approved):
LGTM. Thanks again @Karm

Karm:
@jerboaa , @galderz  Can we proceed? Any further comments?

Karm:
The build is automated now: https://ci.modcluster.io/view/Mandrel/job/mandrel-20.2-windows-build/9/artifact/mandrel-java11-windows-amd64-20.2-SNAPSHOT.zip

galderz (approved):

PR #95, opened 2021-01-22
Jenkins: Fix downstream jobs

zakkak:


zakkak:
I am not sure if this is the right fix though, since I don't see any reference to passing variables/parameters with `/` in https://plugins.jenkins.io/parameterized-trigger/

Karm (approved):

PR #96, opened 2021-01-25
Updates docs on libaries and quickstart

Karm:
Fixes https://github.com/graalvm/mandrel/issues/189

### DIFF
--- a/release-scripts/mandrel-release.java
+++ b/release-scripts/mandrel-release.java
@@ -456,18 +456,19 @@ class MandrelRelease implements Callable<Integer> {
                 "Mandrel's `native-image` depends on the following packages:\n" +
                 "* glibc-devel\n" +
                 "* zlib-devel\n" +
+                "* freetype-devel\n" +
                 "* gcc\n" +
                 "\n" +
                 "On Fedora/CentOS/RHEL they can be installed with:\n" +
                 "```bash\n" +
-                "dnf install glibc-devel zlib-devel gcc libstdc++-static\n" +
+                "dnf install glibc-devel zlib-devel gcc freetype-devel libstdc++-static\n" +
                 "```\n" +
                 "\n" +
                 "Note the package might be called `glibc-static` instead of `libstdc++-static`.\n" +
                 "\n" +
                 "On Ubuntu-like systems with:\n" +
                 "```bash\n" +
-                "apt install gcc zlib1g-dev build-essential\n" +
+                "apt install g++ zlib1g-dev libfreetype6-dev\n" +
                 "```\n" +
                 "\n" +
                 "## Quick start\n" +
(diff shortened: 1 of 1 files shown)


zakkak (approved):

release-scripts/mandrel-release.java
```diff
@@ -456,18 +456,19 @@ private String releaseMainBody(MandrelVersion version, String changelog) {
...
-                "dnf install glibc-devel zlib-devel gcc libstdc++-static\n" +
+                "dnf install glibc-devel zlib-devel gcc freetype-devel libstdc++-static\n" +
                 "```\n" +
                 "\n" +
                 "Note the package might be called `glibc-static` instead of `libstdc++-static`.\n" +
                 "\n" +
                 "On Ubuntu-like systems with:\n" +
                 "```bash\n" +
-                "apt install gcc zlib1g-dev build-essential\n" +
+                "apt install gcc zlib1g-dev build-essential libfreetype6-dev\n" +
```
zakkak, line 471: We might want to reconsider `build-essential` as it's probably installing more than what we need

Karm: @zakkak The reason why I added t was that it was the very thing that brings the correct static glibc and it works on 18.x and 20.x

zakkak: It looks like [build-essential](https://packages.ubuntu.com/bionic/devel/build-essential) installs:
* dpkg-dev (I don't see why we would need that)
* g++
* gcc
* libc6-dev
* make (I don't think we use make anywhere)

We already install `gcc` and `libc6-dev` so it looks like the addition that could possibly fix the issues you observed is `g++` (which judging by the addition of `libstdc++-static` in centos makes sense). Unfortunately we don't have some setup to test this so it will have to be done by hand. I might give it a go later this week.

zakkak: I tested on both 18.04 and 20.04 and the following packages seem enough:
* g++ (this brings gcc as well, which also installs libc6-dev as "recommends")
* zlib1g-dev
* libfreetype6-dev

release-scripts/mandrel-release.java
```diff
@@ -501,7 +502,7 @@ private String releaseMainBody(MandrelVersion version, String changelog) {
                 "One can use the builder image on Windows with Docker Desktop (mind `Resources-> File sharing` settings so as Quarkus project directory is mountable).\n" +
```
aimiebell, line 502: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

zakkak: @aimiebell you are actually pointing to a comment...

PR #106, opened 2021-03-01
Use mx graalvm-version instead of git describe/rev-parse

zakkak:
This way when --version is not given to build.java we get the version
from the suite files just like GraalVM CE does. As a result, we get a
compliant version with org.graalvm.home.Version and avoid issues like
https://github.com/graalvm/mandrel/issues/218

### DIFF
--- a/build.java
+++ b/build.java
@@ -35,6 +35,8 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import javax.management.MBeanException;
+
 import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
 
 public class build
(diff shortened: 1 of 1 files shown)


jerboaa:
I guess this also means we'd have no idea which specific revision was used for building? Do we care about that?

zakkak:
> I guess this also means we'd have no idea which specific revision was used for building? Do we care about that?

I personally never needed that info so far. But that kind of info is probably more useful to @Karm.

Karm:
We need to know which commit hash was used to produce the build, but I don't insist on its location/manifestation as long as it is baked in there somewhere.

zakkak:
org.graalvm.home.Version should be happy with something like 20.3.1.1-dev-1234abcd, let me give it a go

zakkak:
I went for `21.0.0.1-dev337c8550191` to make sure the two dashes in `21.0.0.1-dev-337c8550191` don't become an issue in the future. WDYT?

jerboaa:
> I went for `21.0.0.1-dev337c8550191` to make sure the two dashes in `21.0.0.1-dev-337c8550191` don't become an issue in the future. WDYT?

A scheme like ``21.0.0.1-dev337c8550191` works for me. +1

PR #137, opened 2021-05-14
Housekeeping: Remove cekit builder image and reorganize ansible

### LINKED ISSUE #135: RFC: Thinking of removing the ansible playbook and Cekit modules
The CEKit modules have not been updated for about a year now and AFAIK no one is using them or the ansible playbook (which is more up to date than the CEKit modules)

The idea of having them was to make it easier for developers to setup a dev environment to build Mandrel, but if no one is using them I don't see any value on keeping them around (especially out dated)

\cc @galderz @Karm @jerboaa

zakkak:
I decided to keep the ansible recipes around for now since it's being tested and appears to be easy to maintain 

Closes #135


jerboaa (approved):
LGTM

PR #142, opened 2021-06-03
build.java, insert version string to native-image

Karm:
@zakkak Works for me with Graal master on Linux.
Without it, version string is not inserted and the version ends up being just "dev" instead of expected 

```
native-image 21.2.0-dev bd34f5a08d4 Mandrel Distribution (Java Version 11.0.12-internal+0-adhoc.karm.jdk11u)
```

WDYT?

### DIFF
--- a/build.java
+++ b/build.java
@@ -244,7 +244,7 @@ public class build
         final List<String> lines = Files.readAllLines(nativeImage);
         // This is jamming two sets of parameters in between three sections of command line.
         // It is fragile at best. We should probably just generate the line fresh.
-        final Pattern launcherPattern = Pattern.compile("(.*EnableJVMCI)(.*NativeImage.?)(.*)");
+        final Pattern launcherPattern = Pattern.compile("(.*EnableJVMCI)(.*)");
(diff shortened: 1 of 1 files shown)


build.java
```diff
@@ -232,7 +232,7 @@ private static String getMandrelVersion(Options options, FileSystem fs, Operatin
             List<String> output = os.exec(command, true);
             if (!output.isEmpty())
             {
-                mxVersion += output.get(0);
+                return mxVersion + " " + output.get(0);
```
zakkak, line 235: Please revert this change as it will break compatibility with Quarkus.

Note: in https://github.com/oracle/graal/pull/3329 you can see the latest "version style" we agreed on with the GraalVM team.

zakkak (changes requested):
It looks good to me other than the addition of a space in the version.
Thanks!

zakkak:
For the record, this PR makes mandrel-packaging compatible with the new, after https://github.com/oracle/graal/pull/3382, launcher templates.

zakkak (approved):

PR #170, opened 2021-08-12
JDK17: Adds support for building with https://github.com/openjdk/jdk17u

Karm:
* formatting according to code-formatting-rules.xml
* solves problem of finding correct jars in jdk11 and jdk16 and jdk17 etc. directories
* refactores deleting dirs to Path walker API
* cleans mxbuild dirs before build unless told otherwise

JDK 17
```
$ native-image --version
native-image 21.3.0-devd58f689da739 Mandrel Distribution (Java Version 17-internal+0-adhoc.karm.jdk17u)
```

And JDK 11 compatibility is not broken:
```
$ native-image --version
native-image 21.3.0-devd58f689da739 Mandrel Distribution (Java Version 11.0.12+7)
```

### DIFF
--- a/build.java
+++ b/build.java
@@ -3,10 +3,12 @@ import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.FileSystems;
 import java.nio.file.FileVisitOption;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.PathMatcher;
 import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.StandardOpenOption;
@@ -33,9 +35,8 @@ import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import java.util.stream.Stream;
-
-import javax.management.MBeanException;
 
 import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
 
@@ -49,11 +50,16 @@ public class build
         Check.main();
 
         final Options options = Options.from(Args.read(args));
-
         final FileSystem fs = FileSystem.ofSystem(options);
         final OperatingSystem os = new OperatingSystem();
         final SequentialBuild build = new SequentialBuild(fs, os);
         final Path mandrelRepo = FileSystem.mandrelRepo(options);
+
+        if (!options.skipClean)
+        {
+            logger.debugf("Cleaning mxbuild dirs...");
+            Mx.removeMxbuilds(mandrelRepo);
+        }
 
         logger.debugf("Building...");
         if (options.mandrelVersion == null)
@@ -66,7 +72,7 @@ public class build
 
         logger.debugf("Creating JDK...");
         final Path mandrelHome = FileSystem.mandrelHome(options);
-        FileSystem.deleteDirectoryIfExists(mandrelHome.toFile());
+        FileSystem.deleteRecursively(mandrelHome);
         fs.copyDirectory(os.javaHome(), mandrelHome);
 
         if (!options.skipJava)
@@ -74,7 +80,7 @@ public class build
             logger.debugf("Copy jars...");
             Maven.ARTIFACT_IDS.forEach(artifact ->
             {
-                final String source = mandrelRepo.resolve(Maven.DISTS_PATHS.get(artifact)).toString();
+                final String source = PathFinder.getFirstExisting(Maven.DISTS_PATHS, mandrelRepo, artifact).toString();
                 final String destination = mandrelHome.resolve(Maven.JDK_PATHS.get(artifact)).toString();
                 FileSystem.copy(Path.of(source + ".jar"), Path.of(destination + ".jar"));
                 FileSystem.copy(Path.of(source + ".src.zip"), Path.of(destination + ".src.zip"));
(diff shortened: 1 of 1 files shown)


Karm:
I haven't ran Windows locally this time, so I am curious about the CI result here :)

build.java
```diff
@@ -1498,29 +1554,28 @@ private static void mkdirs(Path path)
...
 
-    public static void deleteDirectoryIfExists(File directory)
+    public static void deleteDirectoryIfExists(Path directory) throws IOException
     {
-        if (!directory.exists())
+        if (Files.notExists(directory))
         {
             return;
         }
-        assert directory.isDirectory();
```
zakkak, line 1507: Can we keep that assertion just in case?

```
directory.toFile().isDirectory();
```

zakkak:
> I haven't ran Windows locally this time, so I am curious about the CI result here :)

I guess we should add JDK 17 builds to the github CI matrix of `mandrel-packaging` and it might be worth having weekly runs on `graalvm/mandrel` as well.
WDYT?

jerboaa:
> > I haven't ran Windows locally this time, so I am curious about the CI result here :)
> 
> I guess we should add JDK 17 builds to the github CI matrix of `mandrel-packaging` and it might be worth having weekly runs on `graalvm/mandrel` as well.
> WDYT?

We don't have static libs available for JDK 17. Oracle doesn't publish them: https://jdk.java.net/17/ We can add it once we have builds for it with static-libs.

build.java
```diff
@@ -771,6 +793,7 @@ static void removeDependencies(Tasks.FileReplace.Effects effects, Function<Path,
         final Path suitePy = Path.of("substratevm", "mx.substratevm", "suite.py");
         final Path path = mandrelRepo.apply(suitePy);
         final List<String> dependencies = Arrays.asList(
+            "jdk.scripting.nashorn@11..14",
```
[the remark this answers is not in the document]
zakkak: I just saw https://github.com/graalvm/mandrel/issues/280

build.java
```diff
@@ -668,6 +675,21 @@ public static Effects noop()
...
     );
 
+    static void removeMxbuilds(Path mandrelRepo) throws IOException
+    {
+        final Path[] mxBuildDirs = new Path[]{
+            Path.of(mandrelRepo.toString(), "compiler", "mxbuild"),
+            Path.of(mandrelRepo.toString(), "java-benchmarks", "mxbuild"),
+            Path.of(mandrelRepo.toString(), "sdk", "mxbuild"),
+            Path.of(mandrelRepo.toString(), "substratevm", "mxbuild"),
+            Path.of(mandrelRepo.toString(), "truffle", "mxbuild")};
```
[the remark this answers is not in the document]
Karm: Addressed  fd9b6a3 

```
DEBUG [build] Cleaning mxbuild dirs...
DEBUG [Mx] Deleting /home/karm/workspaceRH/graal/graal/compiler/mxbuild
DEBUG [Mx] Deleting /home/karm/workspaceRH/graal/graal/java-benchmarks/mxbuild
DEBUG [Mx] Deleting /home/karm/workspaceRH/graal/graal/sdk/mxbuild
DEBUG [Mx] Deleting /home/karm/workspaceRH/graal/graal/substratevm/mxbuild
DEBUG [Mx] Deleting /home/karm/workspaceRH/graal/graal/truffle/mxbuild
DEBUG [build] Building...
```

build.java
Karm: Why? The code does not need such assert anymore as it uses walkFileTree now and it is perfectly capable of receiving a file as an argument.

build.java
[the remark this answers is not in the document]
Karm: > > > I haven't ran Windows locally this time, so I am curious about the CI result here :)
> > 
> > 
> > I guess we should add JDK 17 builds to the github CI matrix of `mandrel-packaging` and it might be worth having weekly runs on `graalvm/mandrel` as well.
> > WDYT?
> 
> We don't have static libs available for JDK 17. Oracle doesn't publish them: https://jdk.java.net/17/ We can add it once we have builds for it with static-libs.

I am building JDK locally and I will set up a CI to do so soon. So instead of building Mandrel with a JDK from Adoptium the job will build JDK first and then build Mandrel with it. That is until there are convenient JDK17 downloadables.

build.java
zakkak: In that case I think the method should be renamed. With the current name I would expect it to do nothing (or complain) if asked to delete "filename" and "filename" is not a directory.

build.java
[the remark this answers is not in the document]
zakkak: Please make sure to build a specific tag of OpenJDK and cache it to avoid rebuilding in every run. I expect this step to take a lot of time.

build.java
```diff
@@ -771,6 +799,7 @@ static void removeDependencies(Tasks.FileReplace.Effects effects, Function<Path,
         final Path suitePy = Path.of("substratevm", "mx.substratevm", "suite.py");
         final Path path = mandrelRepo.apply(suitePy);
         final List<String> dependencies = Arrays.asList(
+            "jdk.scripting.nashorn@11..14",
```
zakkak, line 802: Please add a comment like the following:

```suggestion
            "jdk.scripting.nashorn@11..14",  // work around https://github.com/graalvm/mandrel/issues/280
```

galderz (approved):

build.java
Karm: @zakkak Addressed, squashed.

build.java
Karm: @zakkak Addressed (renamed to `deleteRecursively` as that is what it does). Squashed.

zakkak (approved):
Thanks Karm!

Issue #186, opened 2021-09-27
Two step builds broken on master


zakkak:
The two step build process:

```
# Build the Java bits
${JAVA_HOME}/bin/java -ea build.java --verbose --mx-home ${MX_HOME} --mandrel-repo ${MANDREL_REPO} --mandrel-version "${MANDREL_VERSION}" --skip-native --mandrel-home temp-build 
    
# Build the native bits
./temp-build/bin/java -ea build.java --verbose --mx-home ${MX_HOME} --mandrel-repo ${MANDREL_REPO} --mandrel-version "${MANDREL_VERSION}" --skip-java --archive-suffix tar.gz
```

fails with:

```
Exception in thread "main" java.lang.IllegalAccessError: class com.oracle.graal.pointsto.reports.ReportUtils (in module org.graalvm.nativeimage.pointsto) cannot access class jdk.vm.ci.meta.ResolvedJavaMethod (in module jdk.internal.vm.ci) because module jdk.internal.vm.ci does not export jdk.vm.ci.meta to module org.graalvm.nativeimage.pointsto
	at org.graalvm.nativeimage.pointsto/com.oracle.graal.pointsto.reports.ReportUtils.<clinit>(ReportUtils.java:61)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageOptions.<clinit>(NativeImageOptions.java:182)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageOptions_OptionDescriptors.get(NativeImageOptions_OptionDescriptors.java:17)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageOptions_OptionDescriptors$1.next(NativeImageOptions_OptionDescriptors.java:368)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageOptions_OptionDescriptors$1.next(NativeImageOptions_OptionDescriptors.java:359)
	at jdk.internal.vm.compiler/org.graalvm.compiler.options.OptionKey$Lazy.<clinit>(OptionKey.java:91)
	at jdk.internal.vm.compiler/org.graalvm.compiler.options.OptionKey.getName(OptionKey.java:111)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.<init>(NativeImage.java:193)
[...]

PR #187, opened 2021-09-27
Fix 2 step build process

zakkak:
In the 2 step build process we originally decided to keep the
native-image bash launcher and the macros generation for the
native-image libraries as part of the native build step.

This however is no longer functional as shown in #186 so this patch
moves them to the java build step.

Closes: #186


Karm:
@zakkak Will run Jenkins too....

PR #197, opened 2021-11-05
Remove org.graalvm.libgraal.jni to fix builds after new JVMCI dependency

zakkak:
Mandrel is not using libgraal as a result it doesn't really depend on it.

Closes: https://github.com/graalvm/mandrel/issues/311


jerboaa (approved):
LGTM

PR #209, opened 2021-11-29
(master) Mx builds should use python3 to avoid issues

### LINKED ISSUE #208: Make mandrel-packaging set python version to be used with mx to 3
As python 2 reached its end of life in Jan 2020, we should make sure we no longer use it when building mandrel. `mx` doesn't seem to enforce the use of python3 so if python2 is found in the path it might as well use it.

Related issues:
* https://github.com/graalvm/mx/issues/247
* https://github.com/graalvm/mx/issues/238

galderz:
Closes #208

### DIFF
--- a/build.java
+++ b/build.java
@@ -557,6 +557,12 @@ class EnvVar
         this.name = name;
         this.value = value;
     }
+
+    @Override
+    public String toString()
+    {
+        return name + "=" + value;
+    }
 }
 
 class Tasks
@@ -838,7 +844,7 @@ class Mx
     )
     {
         final Path mx = mxHome.resolve("mx");
-        return Tasks.Exec.of(
+        return execTask(
             Arrays.asList(
                 mx.toString()
                 , options.verbose ? "-V" : ""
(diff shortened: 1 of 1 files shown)


build.java
```diff
@@ -903,6 +909,7 @@ static void build(
             return Tasks.Exec.of(
                 args
                 , mandrelRepo.resolve("substratevm")
+                , new EnvVar("MX_PYTHON", "python3")
```
zakkak, line 912: This only sets the env variable for `mxbuild`.

`mxclean`, `mxversion` and `mandrelVersion` will still run with python 2.

I suggest always adding the variable in `Tasks.Exec.envVars` (even for no mx commands)

zakkak (changes requested):

build.java
galderz: I don't agree with adding the env var to no mx commands. It serves no purpose. I will look into some way to add it to only the mx commands.

galderz:
@zakkak I've updated the PR. I've made such that the env var is set for all mx commands. If you're happy with it I'll backport the latest changes to 21.3.

zakkak (approved):
Thanks @galderz 

LGTM

PR #210, opened 2021-11-29
(21.3) Mx builds should use python3 to avoid issues

### LINKED ISSUE #208: Make mandrel-packaging set python version to be used with mx to 3

galderz:
Closes #208

### DIFF
--- a/build.java
+++ b/build.java
@@ -530,6 +530,12 @@ class EnvVar
         this.name = name;
         this.value = value;
     }
+
+    @Override
+    public String toString()
+    {
+        return name + "=" + value;
+    }
 }
 
 class Tasks
@@ -735,7 +741,7 @@ class Mx
     )
     {
         final Path mx = mxHome.apply(Paths.get("mx"));
-        return Tasks.Exec.of(
+        return execTask(
             Arrays.asList(
                 mx.toString()
                 , options.verbose ? "-V" : ""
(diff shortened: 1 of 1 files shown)


build.java
```diff
@@ -789,6 +795,7 @@ static void build(
             return Tasks.Exec.of(
                 args
                 , mandrelRepo.apply(Path.of("substratevm"))
+                , new EnvVar("MX_PYTHON", "python3")
```
zakkak, line 798: See https://github.com/graalvm/mandrel-packaging/pull/209#discussion_r761781529

galderz:
@zakkak Updated PR.

zakkak (approved):

Karm:
Note: This now requires `python3` to exist, even on a system where `python` actually means Python 3 and there is no Python 2.    So....I am adding

```
mklink C:\Python39\python3.exe C:\Python39\python.exe
```

to make it work now.   Not sure if build.java shouldn't be smarter about it.

zakkak:
Good point.

> Not sure if build.java shouldn't be smarter about it.

It's probably worth checking if `python3` exists before setting it as the `MX_PYTHON`.

PR #218, opened 2022-01-05
Add maven pom files for svm configure tool and native image base

zakkak:
+ Applies https://github.com/graalvm/mandrel-packaging/pull/214 against `master`

  4cc318e778b651a330d9512a6f720afa2de4e34d included the new dependency svm-configure in the build but did not provide pom files for `--maven-install` and `--maven-deploy`

+ Fixes issue with 063842848f3fd6f57ad09423b6f6f364d9ddd8a6 that included the new dependency native-image-base in the build but did not provide pom files for `--maven-install` and `--maven-deploy`
+ Fixes issue with `--maven-install` and `--maven-deploy` caused by cd440c05622ae44064f3f0716a3d6ca3f9c29d29 
+ Adds testing for `--maven-install` and `--maven-deploy`

### DIFF
--- a/resources/assembly/native-image-base/pom.xml
+++ b/resources/assembly/native-image-base/pom.xml
@@ -1,0 +1,82 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.jboss</groupId>
+        <artifactId>jboss-parent</artifactId>
+        <version>36</version>
+    </parent>
+
+    <groupId>org.graalvm.nativeimage</groupId>
+    <artifactId>native-image-base</artifactId>
+    <version>999-ASSEMBLY</version>
+
+    <name>Native Image base</name>
+    <description>Native Image base that can be shared by native image building and pointsto.</description>
+    <url>https://github.com/graalvm/mandrel</url>
+    <developers>
+        <developer>
+            <name>GraalVM Development</name>
+            <email>graalvm-dev@oss.oracle.com</email>
+            <organization>Oracle Corporation</organization>
+            <organizationUrl>http://www.graalvm.org/</organizationUrl>
+        </developer>
+    </developers>
+    <licenses>
+        <license>
+            <name>Universal Permissive License, Version 1.0</name>
+            <url>http://opensource.org/licenses/UPL</url>
+        </license>
+    </licenses>
+    <scm>
+        <connection>scm:git:https://github.com/graalvm/mandrel.git</connection>
+        <developerConnection>scm:git:git@github.com:graalvm/mandrel.git</developerConnection>
+        <url>https://github.com/graalvm/mandrel</url>
+    </scm>
+
+    <dependencies>
+        <dependency>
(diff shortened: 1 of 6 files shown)


resources/assembly/native-image-base/pom.xml
```diff
@@ -0,0 +1,82 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.jboss</groupId>
+        <artifactId>jboss-parent</artifactId>
+        <version>36</version>
+    </parent>
```
jerboaa, line 10: Any reason specific why we depend on `jboss-parent` here? IMO, we should be independent of that parent pom.

resources/assembly/svm-configure/pom.xml
```diff
@@ -0,0 +1,82 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.jboss</groupId>
+        <artifactId>jboss-parent</artifactId>
+        <version>36</version>
+    </parent>
```
jerboaa, line 10: Same.

jerboaa (commented):
I guess this duplicate `artifactId` comment applies to this PR too. Is this standard practise? It seems strange to me to only distinguish between versions (`-ASSEMBLY` suffix).

resources/assembly/native-image-base/pom.xml
zakkak: The new pom files were copied from the existing ones.
@galderz can you please answer @jerboaa's question?

zakkak: FWIW a fix for https://github.com/graalvm/mandrel-packaging/issues/220 would solve such issues.

galderz: I added that dependency early on so that it'd take care of adding all the right metadata to the jar files, which productization required. This parent dependencies have been used throughout middleware productization for many years. If you don't use them, you might need to manually add some metadata to the jar directly in the pom to make productization happy.

galderz: Adding the `jboss-parent` was a quick fix, knowing that I was confident productization would be happy with the metadata it adds to jar files. At the `jar-plugin` level, you can see its effects in [here](https://github.com/jboss/jboss-parent-pom/blob/jboss-parent-36/pom.xml#L428-L456).

galderz: Actually, I think the `assembly-plugin` is relevant here (see [parent contents](https://github.com/jboss/jboss-parent-pom/blob/jboss-parent-36/pom.xml#L205-L233)). This pom then did an override on a couple of properties.

galderz: Btw, there might be other things we got for free by having that parent, e.g. source jar generation...etc. I can't remember of the top of my head, but be ready for some bumps next time we productize the jars if we don't use the parent.

zakkak:
Superseded by https://github.com/graalvm/mandrel-packaging/pull/229

PR #219, opened 2022-01-06
Generate, install, and deploy mandrel-build-wrapper with resources

zakkak:
Enables 2-step build procedure to pass resources generated in the 1st
step (properties and bash launcher) to the 2nd step through a jar
file. This is a requirement for build systems that only allow jar
artifacts to be released in the 1st step.

### DIFF
--- a/build.java
+++ b/build.java
@@ -471,6 +471,7 @@ class Options
 
 class SequentialBuild
 {
+    static final Logger LOG = LogManager.getLogger(SequentialBuild.class);
     final FileSystem fs;
     final OperatingSystem os;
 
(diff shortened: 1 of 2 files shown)


build.java
```diff
@@ -507,17 +487,109 @@ private static String required(String name, Map<String, List<String>> args, bool
...
         }
     }
+
+    private void createJar(Path jarPath, Path... pathsToInclude) throws IOException
+    {
+        Manifest manifest = new Manifest();
+        manifest.getMainAttributes().put(Attributes.Name.MANIFEST_VERSION, "1.0");
+        try (JarOutputStream target = new JarOutputStream(new FileOutputStream(jarPath.toFile()), manifest))
+        {
+            target.setComment("Wrapper jar file for passing native bits to brew");
```
jerboaa, line 527: `s/Wrapper jar file for passing native bits to brew/Wrapper jar file for archiving resources (e.g. native image launcher script)/`

resources/assembly/mandrel-build-wrapper/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+    <scm>
+        <connection>scm:git:https://github.com/graalvm/mandrel.git</connection>
+        <developerConnection>scm:git:git@github.com:graalvm/mandrel.git</developerConnection>
+        <url>https://github.com/graalvm/mandrel</url>
+    </scm>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.graalvm.mandrel</groupId>
+            <artifactId>mandrel-build-wrapper</artifactId>
```
jerboaa, line 43: Could we use a different `artifactId` for the `resource/assembly` file vs. the `resource/release/` file? It's confusing to have the same `artifactId` and only distinguish between versions (at least to me). Have you considered that?

jerboaa (commented):
Looks mostly good.

resources/assembly/mandrel-build-wrapper/pom.xml
zakkak: Not sure. TBH I am not that familiar with the maven assembly, install, and release process. I just followed what we already do for the other artifacts, which was set up by @galderz .

zakkak: FWIW a fix for https://github.com/graalvm/mandrel-packaging/issues/220 would solve such issues.

jerboaa (approved):
Since this is mimicking other pom.xml files we already have, I won't block this PR from getting merged. Perhaps this should get revisited after.

PR #221, opened 2022-01-11
Delegate maven deploy to `mx`

zakkak:
Closes: #220

PR #224, opened 2022-01-17
Add Specification-Version and Implementation-Version to Manifest

zakkak:
Brings back the two properties that got removed from the deployed jars'
manifests with https://github.com/graalvm/mandrel-packaging/pull/221

### DIFF
--- a/.github/workflows/buildJDK.yml
+++ b/.github/workflows/buildJDK.yml
@@ -281,6 +281,7 @@ jobs:
         --mandrel-version "${MANDREL_VERSION}" \
         --skip-native \
         --maven-deploy \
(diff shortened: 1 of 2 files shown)


jerboaa:
Could you comment on why Specification-Version and Implementation-Version in Manifest is needed so that we don't remove those again inadvertently?

zakkak:
> Could you comment on why Specification-Version and Implementation-Version in Manifest is needed so that we don't remove those again inadvertently?

Done

jerboaa (approved):

Karm:
@zakkak Can we merge this so as it is used for the 21.3.1.0 ?

PR #226, opened 2022-01-18
Windows does not allow dirs called aux

Karm:
See https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions


Karm:
THX @zakkak for the idea :+1:

Issue #220, opened 2022-01-11
Use `mx maven-deploy` instead of crafting custom pom files and invoking maven explicitely for deploying mandrel's artifacts 


zakkak:
Right now `build.java` is using [hand-crafted pom files](https://github.com/graalvm/mandrel-packaging/tree/dd2855fc656d963224b2154b7364be605118963f/resources) for [assembling and deploying Mandrel's artifacts](https://github.com/graalvm/mandrel-packaging/blob/dd2855fc656d963224b2154b7364be605118963f/build.java#L1107-L1127).

This, has the following issues:
1. We need to generate new pom files whenever a new artifact becomes available (e.g. 21.3 added svm-configure and 22.0 adds native-image-base).
2. We need to manually track dependencies for the release artifacts.

To avoid those issues we can instead use something like:
```
mx maven-deploy --only SVM --licenses GPLv2-CPE --suppress-javadoc myRepo file:///tmp/maven-repo-mx
```

PR #229, opened 2022-01-24
Delegate maven deploy to `mx` and patch 2-step build process accordingly

zakkak:
Closes: #220


jerboaa (approved):
Yes. Let's keep 21.3 and master ins sync! :)

PR #251, opened 2022-06-09
Don't remove org.graalvm.compiler.truffle.compiler.hotspot.amd64

zakkak:
Works around issue https://github.com/graalvm/mandrel/issues/387 until https://github.com/oracle/graal/pull/4633 gets merged.

Closes https://github.com/graalvm/mandrel/issues/387

### DIFF
--- a/build.java
+++ b/build.java
@@ -1051,7 +1051,7 @@ class Mx
             "\"org.graalvm.compiler.truffle.runtime.hotspot\",",
             "\"org.graalvm.compiler.truffle.runtime.hotspot.java\",",
             "\"org.graalvm.compiler.truffle.runtime.hotspot.libgraal\",",
-            "\"org.graalvm.compiler.truffle.compiler.hotspot.amd64\",",
+            // "\"org.graalvm.compiler.truffle.compiler.hotspot.amd64\",", required by com.oracle.svm.core.meta.ObjectConstantEquality
             "\"org.graalvm.compiler.truffle.compiler.hotspot.aarch64\",",
             "\"org.graalvm.compiler.truffle.jfr\",",
             // "\"truffle:TRUFFLE_API\"", Keep this as there are deps on it


jerboaa:
@zakkak Any ideas why GHA build is failing? Hopefully we don't need this workaround by getting the fix upstream.

zakkak:
> @zakkak Any ideas why GHA build is failing? Hopefully we don't need this workaround by getting the fix upstream.

To my understanding https://github.com/oracle/graal/commit/123490eedd60969c3b903aafa7fe3e970652ec7a#diff-5c2805e0fef0535c28bff8559292cca73a8f81338c2ebe48f1e1039e9db3b68fL1600 removed the explicit dependency on `org.graalvm.compiler.truffle.compiler` (which was satisfying javac before) and now brings it in through `org.graalvm.compiler.truffle.compiler.hotspot.amd64` which we explicitly remove in mandrel-packaging (see https://github.com/oracle/graal/commit/123490eedd60969c3b903aafa7fe3e970652ec7a#diff-5c2805e0fef0535c28bff8559292cca73a8f81338c2ebe48f1e1039e9db3b68fR1856)

Karm:
@zakkak Did this patch help the build for you locally? I might be doing something wrong, but it is the same for me (Windows 10)

```
Building Java module org.graalvm.nativeimage.objectfile (objectfile.jar) from OBJECTFILE
C:\tmp\mandrel\substratevm\src\com.oracle.svm.core\src\com\oracle\svm\core\meta\ObjectConstantEquality.java:28: error: package org.graalvm.compiler.truffle.compiler.nodes does not exist
import org.graalvm.compiler.truffle.compiler.nodes.ObjectLocationIdentity;
                                                  ^
1 error

```

zakkak:
> @zakkak Did this patch help the build for you locally? I might be doing something wrong, but it is the same for me (Windows 10)

Interesting, yes it did work for me locally, but I will need to check on a clean env as well.

Karm:
> > @zakkak Did this patch help the build for you locally? I might be doing something wrong, but it is the same for me (Windows 10)
> 
> Interesting, yes it did work for me locally, but I will need to check on a clean env as well. Scheduled tests on GHA:
> 
>     * linux: https://github.com/graalvm/mandrel/actions/runs/2468710121 heavy_check_mark
> 
>     * windows: https://github.com/graalvm/mandrel/runs/6813751026 hourglass_flowing_sand

The error was on my site. My mandrel source was not clean on Windows. The suites.py were already updated by an older run.

Karm (approved):
Works on my Windows 10 machine.

jerboaa:
@zakkak Why are GHA failing with this? Is this pre-existing?

zakkak:
> @zakkak Why are GHA failing with this? Is this pre-existing?

You mean GHA testing this PR right? These are failing because they are trying to build with the GA JDK, but `graal/master` currently only builds with EA JDK. That's known and OK, not sure if we should change this tests to use EA. It would probably make sense since whenever we change mandrel-packaging is for a future feature release. WDYT?

jerboaa:
> > @zakkak Why are GHA failing with this? Is this pre-existing?
> 
> You mean GHA testing this PR right? These are failing because they are trying to build with the GA JDK, but `graal/master` currently only builds with EA JDK. That's known and OK, not sure if we should change this tests to use EA. It would probably make sense since whenever we change mandrel-packaging is for a future feature release. WDYT?

Ah, OK. Yes, please change it to use EA builds. It would be future proofing things that way.

jerboaa (approved):
Feel free to ship in the interim.

PR #252, opened 2022-06-10
python command is assumed to be Python 3 now

### LINKED ISSUE #217: build.java requires `python3` to exist
As pointed out by @Karm since https://github.com/graalvm/mandrel-packaging/pull/210

> Note: This now requires `python3` to exist, even on a system where python actually means Python 3 and there is no Python 2. So....I am adding
>
> ```
> mklink C:\Python39\python3.exe C:\Python39\python.exe
> ```
>
> to make it work now. Not sure if build.java shouldn't be smarter about it.

We should check if `python3` is available before setting it.

Karm:
fixes #217

It has become very annoying to me to keep symlinking `python3` to `python` on systems, especially those that have *only* Python 3, or e.g. on all new Windows installations. 

I propose to assume `python` is Python 3. The build script bails out very quickly if that is not the case.

### DIFF
--- a/build.java
+++ b/build.java
@@ -3,6 +3,7 @@ import java.io.File;
 import java.io.FileReader;
 import java.io.FileWriter;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.FileSystems;
 import java.nio.file.FileVisitOption;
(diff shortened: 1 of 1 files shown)


zakkak (approved):
LGTM, thanks @Karm !

PR #257, opened 2022-06-16
Drop Ansible setup for building mandrel

zakkak:
This was initially set up to make it easier for new comers to build
Mandrel. However, it looks like it's no longer being used and is now out
of date. Since there is no one actually using it there is no reason
keeping it around and maintaining it.


Karm (approved):
I am not using it.

PR #265, opened 2022-09-09
Clean up dependency removals and further reduce Mandrel's dependencies

zakkak:
Completely drops the TRUFFLE_API dependency (and thus the `TRUFFLE_JCODINGS` one introduced in https://github.com/oracle/graal/commit/5d52020fb4565926a01f870847fe2218aa348e72 as well), with the drawback that it requires some source code patching in `com.oracle.svm.graal`.


jerboaa (approved):
LGTM. We can fix the Windows build as a follow-up.

PR #288, opened 2022-11-02
Refactoring

zakkak:
* replace lambdas with method references
* replace for loop with enhanced for and remove local variable
* Use Scanner(InputStream source, Charset charset) instead of Scanner(InputStream source, String charsetName)


jerboaa (approved):
Seems fine.

PR #292, opened 2022-11-08
Jenkins: Only run tests defined in quarkus/.github/native-tests.json

zakkak:
Limit testing to upstream tests known to work. This is meant to reduce false negatives.

### DIFF
--- a/jenkins/jobs/tests/Constants.groovy
+++ b/jenkins/jobs/tests/Constants.groovy
@@ -39,21 +39,6 @@ class Constants {
                     '2.7.6.Final'
             ]
 
-    static final String QUARKUS_MODULES_TESTS = '' +
-            '!bouncycastle-fips-jsse,' +
-            '!container-image/quarkus-standard-way,' +
-            '!devtools,' +
-            '!google-cloud-functions,' +
-            '!google-cloud-functions-http,' +
-            '!gradle,' +
-            '!kubernetes-client,' +
-            '!kubernetes/maven-invoker-way,' +
-            '!maven,' +
-            '!mongodb-rest-data-panache,' +
-            '!liquibase-mongodb,' +
-            '!mongodb-client,' +
-            '!smallrye-opentracing'
-
     static final String QUARKUS_MODULES_SUBSET_TESTS = '' +
             'bouncycastle-jsse,' +
             'bouncycastle'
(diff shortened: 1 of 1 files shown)


jenkins/jobs/tests/Constants.groovy
```diff
@@ -176,6 +183,11 @@ class Constants {
     git clone --depth 1 --branch %QUARKUS_VERSION% %QUARKUS_REPO%
     cd quarkus
     set "MAVEN_OPTS=-Xmx5g -XX:MaxMetaspaceSize=3g"
+    if "%QUARKUS_MODULES%"=="" (
+        jq -r ".include | map(.\"test-modules\") | join(\", \")" %QUARKUS_REPO%/.github/native-tests.json > modules.txt
+        set /p QUARKUS_MODULES=<modules.txt
+        rm modules.txt
+    )
```
zakkak, line 190: @Karm I am having trouble figuring out how to properly do this. I wan't to check whether `QUARKUS_MODULES` was explicitly set as a parameter and if not then set it using the json file. I would expect the above to work, but haven't managed to confirm :/

zakkak:
@Karm this is a kind reminder for you to have a look at this and provide feedback :)

zakkak:
@Karm any chance you could take a look at this?

zakkak:
ping @Karm

zakkak:
Another ping @Karm

zakkak:
@Karm could you please have a look at this? Is this something worth doing or not?

Karm:
@zakkak Not automated, could differ across very new - very old Q though.
https://github.com/graalvm/mandrel-packaging/blob/master/jenkins/jobs/tests/Constants.groovy#L40

zakkak:
> @zakkak Not automated, could differ across very new - very old Q though.

@karm that's exactly what this is trying to fix. Instead of having a hardcoded list of tests in the Jenkins config, it parses the native-tests.json file that is part of the Quarkus repo we are testing and runs the tests that are defined there (and should work with that Quarkus version). This is how our GHA jobs work for quite some time.

Karm:
> > @zakkak Not automated, could differ across very new - very old Q though.
> 
> @Karm that's exactly what this is trying to fix. Instead of having a hardcoded list of tests in the Jenkins config, it parses the native-tests.json file that is part of the Quarkus repo we are testing and runs the tests that are defined there (and should work with that Quarkus version). This is how our GHA jobs work for quite some time.

Oh. I totally missed the `jq` in the PR. It makes much more sense to me now :) 
The thing is we still want some excludes, e.g. I still removed notoriously unstable or hardly maintainable things from the list.  

I will adapt this branch and test on Windows. I wonder if we could perhaps do it in Groovy right away, at job config build time, rather than relying on system jq later...   That would also mean the list is fixed until mandrel-jobs rebuild.

zakkak:
> The thing is we still want some excludes

I think we can do that by adding our exclusions after the `QUARKUS_MODULES` in the `-pl` argument

> That would also mean the list is fixed until mandrel-jobs rebuild. 

This could be an issue for `main` which is the one expected to change the native-tests.json file. So we might miss some change.

Karm:
I had to rework the Windows logic as `set /p` from a text file silently hits a length limitation, so not all modules were listed there :-)   Yes. That weird.   I use `for` now and it works (tested on Windows):

https://github.com/graalvm/mandrel-packaging/pull/374/files#diff-2c794afcd7598c4375958cb37654db272b51d4565c0a3ec6b839749395d5e67bR224

I adapted your work from this PR. I think we could close this one as soon as it completes...

zakkak:
Superseded by https://github.com/graalvm/mandrel-packaging/pull/374

PR #294, opened 2022-11-08
Jenkins: Skip test jobs if MANDREL_BUILD is not available

zakkak:
When starting a test job manually for a specific MANDREL_BUILD the tests appear to fail in combinations for which the corresponding MANDREL_BUILD doesn't exist. With this change we perform a test and if the build doesn't exist we don't run the job.


Karm:
:pray:

PR #296, opened 2022-11-09
Revert conditionalSteps and introduce combinations filter parameter

zakkak:
It turns out conditionalSteps don't do exactly what I expected, i.e., they don't result in the job being skipped, but instead skip running the tests and mark the job as successful which is misleading (and requires us to accept empty junit reports and artifacts in the publishers).

As a result, I suggest we revert these changes and instead use the [Matrix Combinations Plugin](https://plugins.jenkins.io/matrix-combinations-parameter), which allows one to parameterize the [combinationFilter](https://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.jobs.MatrixJob.combinationFilter) when running a job manually.

This results in the user being prompted with something like the following when building "with Parameters" (which also has more applications than just skipping jobs when a build number is not present for the combination):

image

### DIFF
--- a/jenkins/jobs/builds/mandrel_21_3_linux_build_matrix.groovy
+++ b/jenkins/jobs/builds/mandrel_21_3_linux_build_matrix.groovy
@@ -51,6 +51,7 @@ matrixJob('mandrel-21-3-linux-build-matrix') {
                 '21.3-SNAPSHOT',
                 'It must not contain spaces as it is used in tarball name too.'
         )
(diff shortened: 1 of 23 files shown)


Karm:
Makes sense to me.

I used to do this manually via Axes + parameters + combination filter:
Parameter:
PARAM_X:
Options:
ALL
OPTION_1
OPTION_2

Axes:
OPTIONS:
OPTION_1
OPTION_2

Filter
!(OPTIONS != PARAM_X && PARAM != ALL)

or something to that effect.

The Combinations Filter plugin seem to effectively automate this :+1:

PR #328, opened 2023-03-30
Adjust build scripts for new version scheme

jerboaa:
Depends on upstream version change:
https://github.com/oracle/graal/pull/6356

to use the GraalVM vendor version as we have no way
of knowing the GraalVM code level otherwise.

PR #331, opened 2023-04-06
Set org.graalvm.version for Version.getCurrrent()

jerboaa:
The GraalVM API Version.getCurrent() uses DefaultHomeFinder to get the version String in use and then uses Version.parse() to convert it to a Version object. With the change from #328 that got broken (as the org.graalvm.version) was erronously removed.


jerboaa:
This also depends on an upstream graal fix for https://github.com/oracle/graal/issues/6379 I'll see if I can add the `Version.getCurrent()` reproducer as a smoke test as well.

zakkak (approved):

jerboaa:
Thanks!

PR #337, opened 2023-04-21
Fixes warning on number of JDKs used

Karm:
A small thing, gets rid of:

```
[WARN] There are supposed to be 2 distinct JDK versions used, 
one for JDK 17 and one for JDK 11.This is unexpected: 17.0.7+7
```

when releasing 22.3.

### DIFF
--- a/release-scripts/mandrel-release.java
+++ b/release-scripts/mandrel-release.java
@@ -1100,10 +1100,10 @@ class Release extends ReusableOptions implements Callable<Integer>
             {
                 Log.error("At least one of --windows-job-build-number, --linux-job-build-number must be specified. Terminating.");
             }
-            if (jdkVersionsUsed.size() != 2)
+            if (jdkVersionsUsed.isEmpty() || (jdkVersionsUsed.size() == 1 && version.major != 22) || jdkVersionsUsed.size() > 2)
             {
-                Log.warn("There are supposed to be 2 distinct JDK versions used, one for JDK 17 and one for JDK 11." +
-                    "This is unexpected: " + String.join(",", jdkVersionsUsed));
+                Log.warn("The list of distinct JDK versions used looks weird. " +
+                    "Is this really expected?: " + String.join(",", jdkVersionsUsed));
             }
         }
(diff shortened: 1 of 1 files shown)


release-scripts/mandrel-release.java
```diff
@@ -1100,10 +1100,10 @@ private Set<String> maybeDownloadBuildsAndGetJdkVersions(MandrelVersion version)
             {
                 Log.error("At least one of --windows-job-build-number, --linux-job-build-number must be specified. Terminating.");
             }
-            if (jdkVersionsUsed.size() != 2)
+            if (jdkVersionsUsed.isEmpty() || (jdkVersionsUsed.size() == 1 && version.major != 22) || jdkVersionsUsed.size() > 2)
```
zakkak, line 1103: I am afraid this will get outdated pretty quickly, i.e. 23.0 will use JDK 17 and 20 but 23.1 will probably be JDK 21 only. Maybe warning only when no jdk versions are provided is enough?

Karm: > Maybe warning only when no jdk versions are provided is enough?

I don't like that.  It happened to me that one of the platforms (aarch64?) was not yet uploaded to Temurin's github, so the API call gave our build an old GA JDK.

I missed that fact in Jenkins logs while examining the tests (my error, I don't think I personally would do it again, yet...)

And I caught the error while running this upload as the number of distinct JDK versions was 3 instead of 2.

So I'd like to have this check.

How about encoding the desired JDK x Mandrel versions combinations in a Map globally and using it as a source of truth for this script? Would be easier to update in time than an if statement...

zakkak: > How about encoding the desired JDK x Mandrel versions combinations in a Map globally and using it as a source of truth for this script? Would be easier to update in time than an if statement...

Sounds good.

Karm:
I am publishing 22.3 artifacts from this branch as it works for me. I will revisit the concerns later and tr to get it into master.  Current master doesn't work with the upload as it complains about missing jdk11...

PR #340, opened 2023-05-26
Fix mandrel builds with latest graal master

zakkak:
Closes: https://github.com/graalvm/mandrel/issues/511


zakkak:
These failures are interesting, that's the exact issue I was able to fix locally with this PR :/

jerboaa (approved):
Seems OK.

PR #343, opened 2023-06-06
Added guiding instructions for building Mandrel on macOS

### LINKED ISSUE #341: Packaging does not work on Mac with a locally built base JDK
I build base JDKs from scratch (e.g. see [make target](https://github.com/galderz/dotfiles/blob/main/java17.makefile#L61-L70)). So, assuming a `JAVA_HOME` like the following, the base directory is wrongly calculated:

```
$ ls -al /opt/jdk17u-dev/build/graalbuilder/images/graal-builder-jdk/bin/java
-rwxr-xr-x  1 g  wheel    50K 28 Nov  2022 /opt/jdk17u-dev/build/graalbuilder/images/graal-builder-jdk/bin/java
```

Because it executes `fs.copyDirectory(os.javaHome().getParent().getParent(), mandrelHome);` which means it starts copying `/opt/jdk17u-dev/build/graalbuilder` rather than `/opt/jdk17u-dev/build/graalbuilder/images/graal-builder-jdk`.

I can see 2 ways of fixing this:

1. Some sanitation could be done on the `JAVA_HOME` passed in. For example, only if `$JAVA_HOME/bin/java` is not present on Mac, you can try `$JAVA_HOME/Contents/Home`.

2. Your `JAVA_HOME` should be passed in correctly. If you're on Mac and you use a downloaded base JDK, you have to set `JAVA_HOME=.../Contents/Home`.

Either way, some validation could be done that `bin/java` exists.

galderz:
Closes #341

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,3 +71,30 @@ After the last step you can delete the intermediate JDK with:
 ```shell
 rm -rf /tmp/java_step
 ```
+
+## Building mandrelJDK on a Mac using a custom build of OpenJDK
+
+Mandrel can be built on a Mac using a custom build of OpenJDK that contains static libraries.
+
+If you want to base Mandrel on a custom JDK,
+it's important to build the JDK's macOS bundles first,
+and then extract them into a directory structure that mirrors pre-built JDKs.
+For example:
+
+```shell
+$ make bundles
+$ cd build/<conf-name>/bundles
+$ tar -xzvpf *_bin.tar.gz
+$ tar -xzvpf *_bin-static-libs.tar.gz
+$ ./<jdk-folder>/Contents/Home/bin/java --version
+...
+OpenJDK 64-Bit Server VM (...)
+```
+
+Mandrel should then be built setting `JAVA_HOME` to the macOS-specific java home directory.
+E.g.
+
+```shell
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -71,3 +71,30 @@ After the last step you can delete the intermediate JDK with:
 ```shell
 rm -rf /tmp/java_step
 ```
+
+## Building mandrelJDK on a Mac
+
+Mandrel can be built on a Mac using a supported JDK that contains static libraries.
```
zakkak, line 77:
```suggestion
## Building mandrelJDK on a Mac using a custom build of OpenJDK

Mandrel can be built on a Mac using a custom build of OpenJDK that contains static libraries.
```

zakkak (changes requested):
I think we should make clear that we are talking about custom builds. For pre-built binaries it's pretty much the standard procedure.

galderz:
@zakkak Applied your commit, squashed it and updated PR.

zakkak:
@galderz can you please use my red hat email (in the `Co-authored-by` part of the commit message), or just remove it to get the OCA-bot happy?

galderz:
@zakkak Fixed it

Issue #354, opened 2023-07-20
Add support for local maven deploy


jerboaa:
Our packaging script `build.java` already has an option to deploy to a remote maven repository. We should also add an option to deploy the built artifacts to the local maven repository (usually `~/.m2`).

This would be a step for getting https://github.com/graalvm/mandrel/issues/526 implemented.

PR #355, opened 2023-07-20
Add support for local maven deployment

jerboaa:
Example usage which installs the current version into the local maven repo (`~/.m2` on Linux):

```
$ java -ea ./build.java --maven-deploy-local --mandrel-version $MANDREL_VERSION --verbose --mx-home $MX_HOME --mandrel-repo $MANDREL_REPO --mandrel-home $MANDREL_HOME
```

Closes: #354

### DIFF
--- a/build.java
+++ b/build.java
@@ -1,5 +1,9 @@
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
+
 import java.io.File;
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.FileSystems;
 import java.nio.file.FileVisitOption;
@@ -35,15 +39,13 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
-import static java.nio.charset.StandardCharsets.UTF_8;
-import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
-
 public class build
 {
     static final Logger logger = LogManager.getLogger(build.class);
     public static final boolean IS_WINDOWS = System.getProperty("os.name").matches(".*[Ww]indows.*");
     public static final boolean IS_MAC = System.getProperty("os.name").matches(".*[Mm]ac.*");
     public static final String JDK_VERSION = "jdk" + Runtime.version().feature();
+    public static final String MAVEN_VERSION_FILE = ".maven-version";
     private static final String MANDREL_RELEASE_FILE = "mandrel.release";
 
     public static void main(String... args) throws IOException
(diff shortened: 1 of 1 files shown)


zakkak (approved):
The PR looks good to me but I wonder if it's really necessary.

Can't you achieve local deployment with `--maven-repo-id local --maven-version 3.6.0 --maven-url file://$HOME/.m2/repository`

jerboaa:
Thanks for the review! Don't merge this just yet (still some things coming).

> Can't you achieve local deployment with --maven-repo-id local --maven-version 3.6.0 --maven-url file://$HOME/.m2/repository

It's a `mvn deploy` vs. `mvn install` call. I'd prefer an `install` so as to not even attempt to go remote. Also using the internal `install` logic frees us from trying to re-dediscover the local repo and all that.

jerboaa:
@zakkak Please review again. https://github.com/graalvm/mandrel/issues/526#issuecomment-1645662414 contains a draft as to where this is going.

build.java
```diff
@@ -248,6 +261,32 @@ public static void main(String... args) throws IOException
         }
     }
 
+    private static void writeMavenVersionFile(Options options, Path mandrelHome, String mavenVersion)
+    {
+        if (options.deployLocally)
```
zakkak, line 266: I think moving the check in the caller would be better (keeps clear what `writeMavenVersionFile` does and makes clear when it needs to be done).

build.java
```diff
@@ -248,6 +261,32 @@ public static void main(String... args) throws IOException
...
+    }
+
+    private static String deriveMavenVersion(String mandrelVersion)
+    {
+        if (!mandrelVersion.contains("-dev"))
+        {
+            return mandrelVersion; // use mandrel version verbatim
+        }
+        String baseVers = mandrelVersion.substring(0, mandrelVersion.indexOf("-dev"));
+        return baseVers + ".SNAPSHOT";
```
zakkak, line 287: Nitpick: Quarkus uses `-SNAPSHOT` for snapshot builds despite using `.Final` for releases

zakkak (approved):
Still LGTM, just added two minor comments.

build.java
jerboaa: Sure, but this is for GraalVM produced artefacts, which - I think - uses `.SNAPSHOT` for dev builds. I'd like to keep this in sync.

build.java
jerboaa: Ok, makes sense.

build.java
```diff
@@ -916,6 +964,10 @@ class Mx
     static final List<BuildArgs> DEPLOY_ARTIFACTS_STEPS = List.of(
         BuildArgs.of("--only",
             "GRAAL_SDK," +
+                "sdk:NATIVEIMAGE," +
+                "sdk:COLLECTIONS," +
+                "sdk:POLYGLOT," +
+                "sdk:WORD," +
```
jerboaa, line 970: These four sub-modules need to be added to the deploy artifacts list, since otherwise only GRAAL_SDK gets deployed to the local maven repo and the deps won't get resolved properly (as they're not in the local repo). This is in-line with the `graal-sdk` artefact split.

jerboaa:
This PR depends on #360

zakkak (approved):
LGTM thanks @jerboaa

jerboaa:
CI passes with this in the 23.1 branch. See https://github.com/jerboaa/mandrel-packaging/actions/runs/6099726559

I'm merging this.

PR #358, opened 2023-08-17
Replace graal-sdk with nativeimage, word, and collections

zakkak:
Package was split in https://github.com/oracle/graal/pull/7171

Depends on https://github.com/oracle/graal/pull/7205

### DIFF
--- a/build.java
+++ b/build.java
@@ -878,8 +878,6 @@ class Mx
 
 
         artifacts = Map.ofEntries(
-            new SimpleEntry<>("org.graalvm.sdk:graal-sdk.jar",
-                new Path[]{sdkDistPath.resolve("graal-sdk.jar"), Path.of("lib", "jvmci", "graal-sdk.jar")}),
             new SimpleEntry<>("org.graalvm.nativeimage:svm.jar",
                 new Path[]{substrateDistPath.resolve("svm.jar"), Path.of("lib", "svm", "builder", "svm.jar")}),
             new SimpleEntry<>("org.graalvm.nativeimage:native-image-base.jar",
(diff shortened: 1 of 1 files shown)


jerboaa (commented):
This implicitly will no longer build graal-sdk, which is the only GraalVM dependency in Quarkus. We need to carefully look at what dependencies we need to replace this with on the Quarkus end. Also affects graal-sdk not being built with `--java-only` which affects the product build.

zakkak:
> This implicitly will no longer build graal-sdk, which is the only GraalVM dependency in Quarkus. We need to carefully look at what dependencies we need to replace this with on the Quarkus end. Also affects graal-sdk not being built with `--java-only` which affects the product build.

Good point, so before moving forward with this we need to see what the plans for the upstream graal-sdk maven artifact are. If it's indeed going to keep getting published as is we should not remove it from our builds.

jerboaa:
I'm thinking to go with this one for now and keep a close eye out on https://github.com/oracle/graal/discussions/7224 and do all necessary preparations to move to the new API maven artefact.

jerboaa:
On second thought, we might need to go with #359 after all so as to build `graal-sdk` which will be the dep Quarkus is currently using. It cannot be changed easily, since Quarkus wants to work with older GraalVM releases as well. So we might want to change to a slimmer build a few releases in the future only.

jerboaa:
This PR has the same problem when testing `GRAAL_SDK` local deployment as #359. See https://github.com/graalvm/mandrel-packaging/pull/359#issuecomment-1686653281

zakkak:
superseded by #360

PR #359, opened 2023-08-17
Explicitly build graal-sdk and add nativeimage, word and collections

zakkak:
New packages introduced in oracle/graal#7171

Depends on https://github.com/oracle/graal/pull/7205

Supersedes https://github.com/graalvm/mandrel-packaging/pull/358 (?)

PR #360, opened 2023-08-21
Account for split of graal-sdk artefact

jerboaa:
Better fix to #359 which allows for local deployment of those
modules `graal-sdk` now depends on. See #355.


zakkak (approved):
LGTM, thanks!

PR #361, opened 2023-09-05
Temporarily move to building 23.1

jerboaa:
Graal `master` branch is currently not buildable (in the Mandrel config) and is, thus, not helpful for CI runs. In order to avoid breakages I propose to move to 23.1 temporarily.

This patch also does the following (in separate commits):

- Move to JDK 21 for building (23.1 is JDK 21 only)
- Disable the Version test which uses org.graalvm.home.Version and related classes now in `polyglot.jar` currently not part of Mandrel 23.1
- Also run `native-image --version` after the launcher macro build since we can catch bugs that way like https://github.com/graalvm/mandrel/issues/556

### DIFF
--- a/.github/workflows/buildJDK.yml
+++ b/.github/workflows/buildJDK.yml
@@ -39,7 +39,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        mandrel-ref: [graal/master]
+        mandrel-ref: [mandrel/23.1]
     steps:
     - uses: actions/checkout@v3
     - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


jerboaa:
Switching to JDK 21 will be needed too...

jerboaa:
https://github.com/graalvm/mandrel-packaging/pull/361/commits/a7f770dc54ce4174a583cd5bad49700853c03052 should be easily revertable once we have 24.0/master working again.

jerboaa:
OK. Replaced by https://github.com/graalvm/mandrel-packaging/pull/363

jerboaa:
> For building/testing 23.1 builds we should use https://github.com/graalvm/mandrel-packaging/tree/23.1
> IMO if we need a change for 23.1 that we probably want in master as well we should open a PR against 23.1

PR against 23.1 is here:
https://github.com/graalvm/mandrel-packaging/pull/362

PR #362, opened 2023-09-06
Change CI workflows to use 23.1

jerboaa:
Fix CI for 23.1 branch.

This patch does the following (in separate commits):

- Move to JDK 21 for building (23.1 is JDK 21 only)
- Disable the Version test which uses org.graalvm.home.Version and related classes now in `polyglot.jar` currently not part of Mandrel 23.1
- Also run `native-image --version` after the launcher macro build since we can catch bugs that way like https://github.com/graalvm/mandrel/issues/556

### DIFF
--- a/.github/workflows/buildJDK.yml
+++ b/.github/workflows/buildJDK.yml
@@ -39,7 +39,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        mandrel-ref: [graal/master]
+        mandrel-ref: [mandrel/23.1]
     steps:
     - uses: actions/checkout@v3
     - uses: actions/checkout@v3
@@ -140,7 +140,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        mandrel-ref: [graal/master]
+        mandrel-ref: [mandrel/23.1]
     steps:
       - uses: actions/checkout@v3
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


zakkak (approved):

zakkak:
The 2 step build fails with:

```
AttributeError: 'Namespace' object has no attribute 'generate_javadoc'
```

Could it be that we are using the wrong mx version?

jerboaa:
> The 2 step build fails with:
> 
> ```
> AttributeError: 'Namespace' object has no attribute 'generate_javadoc'
> ```
> 
> Could it be that we are using the wrong mx version? i.e. mx `master` might not be compatible with 23.1

Possibly. It passed yesterday, though (with `mx` revision `c86b9e88a23ec263d9f19b754b29d30dd18272da`). See:
https://github.com/graalvm/mandrel-packaging/actions/runs/6087706591

Let me try with it.

zakkak:
I suggest pinning the mx version to 6.42.0, see https://github.com/graalvm/mandrel/blob/mandrel/23.1/common.json#L7

jerboaa:
> I suggest pinning the mx version to 6.42.0 (or possibly get it from common.json, although I recall having some difficulty in doing so), see https://github.com/graalvm/mandrel/blob/mandrel/23.1/common.json#L7

Sure that works too. I'll let the run with `6.45.0` complete and then possibly tune it down to `6.42.0`

jerboaa:
@zakkak CI is happy. Do you want me to use `6.42.0` instead or is `6.45.0` good enough?

zakkak:
I am OK with either.

jerboaa:
OK, merging as is then. Thanks!

PR #375, opened 2023-10-25
JAVA_HOME is no longer GRAALVM_HOME, defaults to Temurin 17.

Karm:


Karm:
Brings regression for commands using native-agent:
```
Command: [java, -Djava.awt.headless=true, -agentlib:native-image-agent=config-output-dir=src/main/resources/META-INF/native-image, -jar, target/imageio.jar]
Error occurred during initialization of VM
Could not find agent library native-image-agent on the library path, with error: libnative-image-agent.so: cannot open shared object file: No such file or directory
```

Issue #382, opened 2023-11-20
[CI] MacOS GHA build is currently broken


jerboaa:
MacOS builds run via GHA are currently broken. They fail with:

```
The following internal mx symbols were accessed: mx_benchmark._mx_benchmark_usage_example, mx._mx_path
  File "/Users/runner/work/mandrel-packaging/mandrel-packaging/mx/src/mx/_impl/mx.py", line 2090, in _load
    self._load_extensions()
  File "/Users/runner/work/mandrel-packaging/mandrel-packaging/mx/src/mx/_impl/mx.py", line 2587, in _load_extensions
    mod = __import__(extensionsName)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/runner/work/mandrel-packaging/mandrel-packaging/mandrel/substratevm/mx.substratevm/mx_substratevm.py", line 31, in <module>
    from distutils.dir_util import mkpath  # pylint: disable=no-name-in-module
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'distutils'
Exception in thread "main" java.lang.RuntimeException: java.lang.RuntimeException: Failed, exit code: 1
	at OperatingSystem.exec(build.java:1821)
	at SequentialBuild.lambda$build$0(build.java:638)
	at Mx.build(build.java:1029)
	at SequentialBuild.build(build.java:640)
	at build.main(build.java:83)
Caused by: java.lang.RuntimeException: Failed, exit code: 1
	at OperatingSystem.exec(build.java:1810)
	at SequentialBuild.lambda$build$0(build.java:638)
	at Mx.build(build.java:1029)
	at SequentialBuild.build(build.java:640)
	at build.main(build.java:83)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at jdk.compiler/com.sun.tools.javac.launcher.Main.execute(Main.java:484)
	at jdk.compiler/com.sun.tools.javac.launcher.Main.run(Main.java:208)
	at jdk.compiler/com.sun.tools.javac.launcher.Main.main(Main.java:135)
Error: Process completed with exit code 1.
```

jerboaa:
[...]

PR #383, opened 2023-11-21
[CI] Pin python version to 3.10

zakkak:
`mx` uses `distutils` which is no longer available in python 3.12 which
is the default in `macos-latest`

Use 3.10 instead in all platforms for consistency. The minimum supported
version by `mx` is 3.9 see https://github.com/graalvm/mx/issues/249.

Closes https://github.com/graalvm/mandrel-packaging/issues/382


jerboaa (approved):
OK.

PR #391, opened 2024-01-15
[23.1] Fix preview build

jerboaa:
Currently using `--enable-preview` is broken for Mandrel builds, since it doesn't install `svm-foreign.jar` in the right directory. This patch adds the relevant dependency (`SVM_FOREIGN`) and then installs it.

Closes: https://github.com/graalvm/mandrel/issues/658


jerboaa:
I'll forward-port this to 24.0 and master once CI passes.

jerboaa:
Thanks for the review!

PR #398, opened 2024-01-26
Only build and include svm-foreign when building with JDK > 21

zakkak:
Since https://github.com/oracle/graal/pull/7980 the foreign API is only
available for JDK >= 22 builds.

Fixes error message:

```
dependency named SVM_FOREIGN was removed:
  distribution SVM_FOREIGN was removed as all its dependencies were removed:
    project com.oracle.svm.hosted.foreign was removed as JDK 22 is not available
```

When trying to build with JDK 21.

If accepted I will backport it to 24.0 as well.

### DIFF
--- a/build.java
+++ b/build.java
@@ -138,10 +138,13 @@ public class build
                 FileSystem.copy(mandrelRepo.resolve(
                     Path.of("sdk", "mxbuild", PLATFORM, "native-image.exe.image-bash", "native-image.export-list")), nativeImageExport);
             }
-            logger.debugf("Copy svm-preview...");
-            final Path svmForeign = mandrelJavaHome.resolve(Path.of("lib", "svm-preview", "builder", "svm-foreign.jar"));
-            final Path svmForeignSource = PathFinder.getFirstExisting(mandrelRepo.resolve(Path.of("substratevm", "mxbuild")).toString(), "svm-foreign.jar");
-            FileSystem.copy(svmForeignSource, svmForeign);
+            if (Runtime.version().feature() > 21)
+            {
+                logger.debugf("Copy svm-preview...");
+                final Path svmForeign = mandrelJavaHome.resolve(Path.of("lib", "svm-preview", "builder", "svm-foreign.jar"));
+                final Path svmForeignSource = PathFinder.getFirstExisting(mandrelRepo.resolve(Path.of("substratevm", "mxbuild")).toString(), "svm-foreign.jar");
+                FileSystem.copy(svmForeignSource, svmForeign);
+            }
         }
 
         if (!options.skipNative)
@@ -855,7 +858,7 @@ class Mx
         Pattern.compile("\"version\"\\s*:\\s*\"([0-9.]*)\"");
 
     static final List<BuildArgs> BUILD_JAVA_STEPS = List.of(
-        BuildArgs.of("--no-native", "--dependencies", "SVM,SVM_FOREIGN,GRAAL_SDK,SVM_DRIVER,SVM_AGENT,SVM_DIAGNOSTICS_AGENT")
+        BuildArgs.of("--no-native", "--dependencies", "SVM,GRAAL_SDK,SVM_DRIVER,SVM_AGENT,SVM_DIAGNOSTICS_AGENT" + (Runtime.version().feature() > 21 ? ",SVM_FOREIGN" : ""))
         , BuildArgs.of("--only",
             build.IS_WINDOWS ?
                 "native-image.exe.image-bash," +


build.java
```diff
@@ -854,20 +857,34 @@ class Mx
     private static final Pattern DEPENDENCY_VERSION_PATTERN =
         Pattern.compile("\"version\"\\s*:\\s*\"([0-9.]*)\"");
 
-    static final List<BuildArgs> BUILD_JAVA_STEPS = List.of(
-        BuildArgs.of("--no-native", "--dependencies", "SVM,SVM_FOREIGN,GRAAL_SDK,SVM_DRIVER,SVM_AGENT,SVM_DIAGNOSTICS_AGENT")
```
jerboaa, line 858: Could we use this instead?

```
BuildArgs.of("--no-native", "--dependencies", "SVM,GRAAL_SDK,SVM_DRIVER,SVM_AGENT,SVM_DIAGNOSTICS_AGENT" + (Runtime.version().feature() > 21: ",SVM_FOREIGN" ? "");
```

jerboaa (commented):
I'm not convinced we need JDK 21 build support, but your mileage may vary.

jerboaa:
Also note that IMO it's an (upstream) bug to add attempt to add `svm-foreign.jar` if `--enable-preview` is specified. I'm pretty sure a Graal master build with labsjdk 21 and `--enable-preview` will fail since in that case the native image generator unconditionally attempts to add the dir containing `svm-foreign.jar` (not present) to the class/module path.

build.java
zakkak: Sure

zakkak:
> Also note that IMO it's an (upstream) bug to add attempt to add `svm-foreign.jar` if `--enable-preview` is specified. I'm pretty sure a Graal master build with labsjdk 21 and `--enable-preview` will fail since in that case the native image generator unconditionally attempts to add the dir containing `svm-foreign.jar` (not present) to the class/module path.

I think that's fixed in https://github.com/oracle/graal/pull/7980

jerboaa (approved):

jerboaa:
CI fails due to too new JDK 23:
```
Error: Static modifier mismatch: private void com.oracle.svm.core.thread.Target_java_lang_VirtualThread.notifyJvmtiHideFrames(boolean), private static native void java.lang.VirtualThread.notifyJvmtiHideFrames(boolean)
```

Fixed once this PR merges: https://github.com/oracle/graal/pull/8262

jerboaa:
> > Also note that IMO it's an (upstream) bug to add attempt to add `svm-foreign.jar` if `--enable-preview` is specified. I'm pretty sure a Graal master build with labsjdk 21 and `--enable-preview` will fail since in that case the native image generator unconditionally attempts to add the dir containing `svm-foreign.jar` (not present) to the class/module path.
> 
> I think that's fixed in [oracle/graal#7980](https://github.com/oracle/graal/pull/7980)

You are right. Thinking some more about this, we don't need the `SVM_FOREIGN` dependency on JDK 23 either. Not sure why the build doesn't outright fail...

zakkak:
> we don't need the SVM_FOREIGN dependency on JDK 23 either. Not sure why the build doesn't outright fail...

Why is that? All the checks in GraalVM are for `>= 22`.

build.java
```diff
@@ -138,10 +138,13 @@ public static void main(String... args) throws IOException
...
                     Path.of("sdk", "mxbuild", PLATFORM, "native-image.exe.image-bash", "native-image.export-list")), nativeImageExport);
             }
-            logger.debugf("Copy svm-preview...");
-            final Path svmForeign = mandrelJavaHome.resolve(Path.of("lib", "svm-preview", "builder", "svm-foreign.jar"));
-            final Path svmForeignSource = PathFinder.getFirstExisting(mandrelRepo.resolve(Path.of("substratevm", "mxbuild")).toString(), "svm-foreign.jar");
-            FileSystem.copy(svmForeignSource, svmForeign);
+            if (Runtime.version().feature() > 21)
+            {
+                logger.debugf("Copy svm-preview...");
+                final Path svmForeign = mandrelJavaHome.resolve(Path.of("lib", "svm-preview", "builder", "svm-foreign.jar"));
```
jerboaa, line 144: `svm-preview` folder is wrong as a target for JDK 22+. See https://github.com/graalvm/mandrel/issues/678 for some details.

jerboaa (changes requested):
It makes little sense to me to try to keep JDK 21 compatibility in `master`. The entire JEP 454 story in GraalVM seems not 100% ready yet. See https://github.com/graalvm/mandrel/issues/678

jerboaa:
> > we don't need the SVM_FOREIGN dependency on JDK 23 either. Not sure why the build doesn't outright fail...
> 
> Why is that? All the checks in GraalVM are for `>= 22`.

Sorry, I was wrong. However, there is a different problem with getting `svm-foreign.jar` properly installed for JDK 23. I'll take a look.

jerboaa:
> > > we don't need the SVM_FOREIGN dependency on JDK 23 either. Not sure why the build doesn't outright fail...
> > 
> > 
> > Why is that? All the checks in GraalVM are for `>= 22`.
> 
> Sorry, I was wrong. However, there is a different problem with getting `svm-foreign.jar` properly installed for JDK 23. I'll take a look.

See #399, which should fix the `svm-foreign.jar` installation.

zakkak:
> It makes little sense to me to try to keep JDK 21 compatibility in master. The entire JEP 454 story in GraalVM seems not 100% ready yet. See https://github.com/graalvm/mandrel/issues/678

The primary reason for opening this was to prevent build failures when trying to build with JDK 21 so that we can see the actual blocker/issue which is https://github.com/graalvm/mandrel/issues/598

Closing this for now as there are no plans to build `master` with JDK 21, see https://github.com/graalvm/mandrel/issues/598#issuecomment-1912118196

PR #399, opened 2024-02-01
[JDK 22/23] Fix installation/copy of svm-foreign.jar

jerboaa:
The copying for `svm-foreign.jar` in `master` is broken it copied the directory `substratevm/mxbuild/dists/jdk23` to file `$MANDREL_HOME/lib/svm/svm-preview/svm-foreign.jar` which is broken for two reasons:

1. It's not actually a jar that gets intalled, but a directory
2. It installs into the wrong (old JDK 21) installation path: `lib/svm/builder/svm-preview` over JDK 22+ location next to `svm.jar` in `lib/svm/builder` directory.

This patch fixes this and also copies the corresponding `src.zip` which we do for other artefacts as well.

Thoughts?

### DIFF
--- a/build.java
+++ b/build.java
@@ -138,10 +138,6 @@ public class build
                 FileSystem.copy(mandrelRepo.resolve(
                     Path.of("sdk", "mxbuild", PLATFORM, "native-image.exe.image-bash", "native-image.export-list")), nativeImageExport);
             }
-            logger.debugf("Copy svm-preview...");
-            final Path svmForeign = mandrelJavaHome.resolve(Path.of("lib", "svm-preview", "builder", "svm-foreign.jar"));
-            final Path svmForeignSource = PathFinder.getFirstExisting(mandrelRepo.resolve(Path.of("substratevm", "mxbuild")).toString(), "svm-foreign.jar");
-            FileSystem.copy(svmForeignSource, svmForeign);
         }
 
         if (!options.skipNative)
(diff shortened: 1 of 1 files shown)


zakkak (commented):
@jerboaa why not add svm-foreign.jar to `Mx.artifacts` and let the "Copy jars" step in line 112 take care of it as it does for the rest?

jerboaa:
> @jerboaa why not add svm-foreign.jar to `Mx.artifacts` and let the "Copy jars" step in line 112 take care of it as it does for the rest?

Good point. Do you know which artefact coordinates I need to use?

zakkak:
> Good point. Do you know which artefact coordinates I need to use?

I'd say `org.graalvm.nativeimage:svm-foreign.jar`, but it shouldn't really matter as we are not publishing it as a maven artifact, i.e. the coordinates won't be used except in a potential error message if the jar file is not found.

jerboaa:
> > Good point. Do you know which artefact coordinates I need to use?
> 
> I'd say `org.graalvm.nativeimage:svm-foreign.jar`, but it shouldn't really matter as we are not publishing it as a maven artifact, i.e. the coordinates won't be used except in a potential error message if the jar file is not found.

OK

jerboaa:
@zakkak Done now. Please have another look.

zakkak (approved):
LGTM, thanks @jerboaa

PR #400, opened 2024-02-01
[JDK 22/23] Fix installation/copy of svm-foreign.jar

jerboaa:
Same as #399 but for the 24.0 (JDK 22) branch. The difference being that the copy finds the right `svm-foreign.jar`, but still copies it to the wrong location. The finding-the-wrong source to copy from is a master only bug.

Thoughts?


jerboaa:
@zakkak PTAL. Same as for master, but for the JDK 22/24.0 branch.

zakkak (approved):

Issue #403, opened 2024-02-13
Revert "Set MaxRAM and MaxRAMPercentage in launcher for native-image driver" https://github.com/graalvm/mandrel-packaging/pull/366


zakkak:
Revert https://github.com/graalvm/mandrel-packaging/pull/366 if/when https://github.com/oracle/graal/pull/8384 gets merged.

jerboaa:
https://github.com/oracle/graal/pull/8384 is merged in 24.1. It's not the 100% equivalent, but it'll probably do.

PR #412, opened 2024-03-27
Revert "Set MaxRAM and MaxRAMPercentage in launcher for native-image driver"

zakkak:
This reverts commit 7fc9374f936d64386ab6363d5dabe7a3f8d63136.

Closes https://github.com/graalvm/mandrel-packaging/issues/403


jerboaa (approved):
OK, although I still don't like the upstream solution. Sigh.

zakkak:
I understand, but it's still better than nothing and we should better not diverge. So it is what it is.

PR #415, opened 2024-05-27
Test mac-os builds both on Intel and Apple M

zakkak:
`macos-13` is the latest intel-based runner, see https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

### DIFF
--- a/.github/workflows/buildJDK.yml
+++ b/.github/workflows/buildJDK.yml
@@ -140,11 +140,12 @@ jobs:
         path: mandrel-java23-linux-amd64.tarxz
 
   build-and-test-on-mac:
-    name: MacOS Build and test ${{ matrix.mandrel-ref }} branch/tag
-    runs-on: macos-latest
+    name: ${{ matrix.os }} Build and test ${{ matrix.mandrel-ref }} branch/tag
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
+        os: [macos-latest, macos-13]
         mandrel-ref: [graal/master]
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


jerboaa:
Looks like we need something like https://github.com/graalvm/mandrel/pull/735 as well.

zakkak:
> Looks like we need something like [graalvm/mandrel#735](https://github.com/graalvm/mandrel/pull/735) as well.

Done in https://github.com/graalvm/mandrel-packaging/pull/416

Karm:
@zakkak I had to rename lib/static/darwin-arm64  to lib/static/darwin-aarch64 too :man_shrugging:

zakkak:
> @zakkak I had to rename lib/static/darwin-arm64 to lib/static/darwin-aarch64 too 🤷‍♂️
> 
> working with 23.1 branch

That sounds like an upstream bug, shouldn't upstream use the same naming as OpenJDK?

Also any idea why the smoke tests seem to work without this renaming in this case?

Karm:
@zakkak 

> > @zakkak I had to rename lib/static/darwin-arm64 to lib/static/darwin-aarch64 too 🤷‍♂️
> > working with 23.1 branch
> 
> That sounds like an upstream bug, shouldn't upstream use the same naming as OpenJDK?
> 

I have seen these in the wild, JNI Java libs, projects, JDK, GraalVM, uname, arch etc.:

```
osx-aarch64
osx-arm
osx-arm64
mac-aarch64
mac-arm
mac-arm64
macos-aarch64
macos-arm
macos-arm64
darwin-aarch64
darwin-arm
darwin-arm64
```

The ISA is apparently interchangeably called both Aarch64 and ARM64. It seems that MacOS pretty consistently reports "arm64" though. Both via `file` command examining Mach-0 executables and with `uname` and `arch`.
The kernel is called "Darwin" in `uname`, but  `$(sw_vers -productName) ` tells you "macOS". 

Temurin 21 uses "lib/static/darwin-arm64", although it names its tarballs "aarch64_mac".
GraalVM CE uses "lib/static/darwin-aarch64" and they call their tarballs "macos-aarch64". 

> Also any idea why the smoke tests seem to work without this renaming in this case?

My hunch is that this naming mayhem is so prevalent a problem that the GHA runners have some kind of aliasing in place to mitigate failing on such trivial naming steps? When I am done with the baremetal ones, I'll take a look.

Karm:
@jerboaa 
@zakkak 
What we need to agree on is whether we copy what Temurin does, both in naming the lib dir and in naming the tarballs,
or whether we copy what GraalVM CE does.

The one who suggests we solve this by creating a third naming scheme wins a t-shirt :)

Karm (approved):

zakkak:
> @jerboaa @zakkak What we need to agree on is whether we copy what Temurin does, both in naming the lib dir and in naming the tarballs, or whether we copy what GraalVM CE does.
> 
> The one who suggests we solve this by creating a third naming scheme wins a t-shirt :)

I would love to see the print on that T-shirt but I will be logical and not suggest a third naming scheme :)

Although I would like to go with the Temurin naming scheme I am afraid the easiest path is to go with the GraalVM CE naming scheme as I expect `mx` to expect/generate the names and I don't think it's worth the effort to try and detect where and how that happens. Furthermore, less deviation from upstream seems better to me in general (we could always also ask upstream for the reasoning behind the different naming).

zakkak:
> Also any idea why the smoke tests seem to work without this renaming in this case?

It turns out it's still building for x86_64

```
C compiler: cc (apple, x86_64, 15.0.0)
```

https://github.com/graalvm/mandrel-packaging/actions/runs/9256122608/job/25461451554?pr=415#step:9:520

Switching to draft... I need to change more things (like the openjdk we download) and make sure rosetta is not taking over and emulating things.

Karm:
@zakkak Just to confirm, the build I tried was on a system without Rosetta, so it works, but for the naming....

zakkak:
> @zakkak Just to confirm, the build I tried was on a system without Rosetta, so it works, but for the naming....

Yes, I fixed it in the github action now as well, but it looks like this renaming needs to be done by `build.java` instead. I'll give it some more thought and try to understand why GraalVM CE uses a different naming scheme.

.github/workflows/buildJDK.yml
```diff
@@ -194,12 +208,14 @@ jobs:
...
+        env:
+          ARCH: ${{ steps.arch.outputs.ARCH }}
         run: |
           export JAVA_HOME=${MAC_JAVA_HOME}
           ${MAC_JAVA_HOME}/bin/java -ea build.java --mx-home ${MX_HOME} --mandrel-repo ${MANDREL_REPO} --mandrel-version "${MANDREL_VERSION}" --archive-suffix tar.gz
           export MANDREL_VERSION_UNTIL_SPACE="$( echo ${MANDREL_VERSION} | sed -e 's/\([^ ]*\).*/\1/;t' )"
-          export ARCHIVE_NAME="mandrel-java23-darwin-amd64-${MANDREL_VERSION_UNTIL_SPACE}.tar.gz"
-          mv ${ARCHIVE_NAME} mandrel-java23-darwin-amd64.tar.gz
+          export ARCHIVE_NAME="mandrel-java23-darwin-${ARCH}-${MANDREL_VERSION_UNTIL_SPACE}.tar.gz"
+          mv ${ARCHIVE_NAME} mandrel-java23-darwin-${ARCH}.tar.gz
```
jerboaa, line 218: This seems to cause (on mac x86_64):

```
INFO [build] Creating Archive mandrel-java23-darwin-amd64-10.9.8.7-dev.tar.gz
 mv: rename mandrel-java23-darwin-x64-10.9.8.7-dev.tar.gz to mandrel-java23-darwin-x64.tar.gz: No such file or directory
Error: Process completed with exit code 1.
```

zakkak:
> That sounds like an upstream bug, shouldn't upstream use the same naming as OpenJDK?

So it looks like a Temurin issue after all. I opened https://github.com/adoptium/temurin-build/pull/3827 to fix this.

PR #435, opened 2024-07-10
Better handling of MANIFEST.MF attributes

jerboaa:
Previously, the build script would add these two attributes irrespective whether or not they'd exist in the Graal artefacts:

```
  Specification-Version
  Implementation-Version
```

With this patch, only updates to the Implementation-Version attribute are done based on the maven version in use. Specification-Version is only added if it's not yet there in a jar or source jar file.

Closes: https://github.com/Karm/mandrel-integration-tests/issues/270

### DIFF
--- a/build.java
+++ b/build.java
@@ -3,6 +3,8 @@ import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.FileSystems;
@@ -33,6 +35,8 @@ import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
+import java.util.jar.Attributes;
+import java.util.jar.Manifest;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


build.java
```diff
@@ -693,15 +699,31 @@ private BiConsumer<String, Path[]> amendOrCreateManifest(Options options, Tasks.
...
+        attributes.putValue(IMPLEMENTATION_VERSION_NAME, options.mavenVersion);
+        String specVers = attributes.getValue(SPECIFICATION_VERSION_NAME);
+        if (specVers == null) {
+            // Only add Specification-Version if not already present
+            attributes.putValue(SPECIFICATION_VERSION_NAME, "0.0");
+        }
+        String manifestVers = attributes.getValue(Attributes.Name.MANIFEST_VERSION);
+        if (manifestVers == null) {
+            attributes.putValue("Manifest-Version", "1.0");
+        }
```
jerboaa, line 721: The `Manifest-Version` is added so as to make `mf.write()` API work which requires at least that attribute being present.

zakkak (approved):
LGTM. My only hesitation is about the Specification-Version not being the same.

jerboaa:
> LGTM. My only hesitation is about the `Specification-Version` not being the same with the `Implementation-Version`.

Thanks for the review! This is pre-existing. We were adding `Specification-Version: 0.0` unconditionally before while setting `"Implementation-Version: " + options.mavenVersion`. The comment of the method suggests it wasn't important. Also, what would we set the specification version to if it's not there in the original jar?

PR #438, opened 2024-07-17
[24.0] Backport: Better handling of MANIFEST.MF attributes

zakkak:
Backport of https://github.com/graalvm/mandrel-packaging/pull/435

> Previously, the build script would add these two attributes irrespective
> whether or not they'd exist in the Graal artefacts:
> 
>   Specification-Version
>   Implementation-Version
> 
> With this patch, only updates to the Implementation-Version attribute
> are done based on the maven version in use. Specification-Version is
> only added if it's not yet there in a jar or source jar file.

(cherry picked from commit 48c154e09da76394c09f3ed9b1ae03ed699d0df5)

### DIFF
--- a/build.java
+++ b/build.java
@@ -3,6 +3,8 @@ import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.FileSystems;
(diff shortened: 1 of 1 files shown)


zakkak:
@jerboaa In 24.0 we need it due to https://github.com/oracle/graal/commit/d2bba8feea48fc6e224c43803f4bfe65099d7ecd and https://github.com/graalvm/mx/commit/5eb6d6f73d16a37f52c9d69815af479523513bdd

24.1 is still using mx 7.25.5 which doesn't include https://github.com/graalvm/mx/commit/bafad626087bccdc7cb1032cca68a5bcddff1ff2 so for the time being we don't need the backport (except for consistency).

jerboaa:
> 24.1 is still using mx 7.25.5 which doesn't include [graalvm/mx@bafad62](https://github.com/graalvm/mx/commit/bafad626087bccdc7cb1032cca68a5bcddff1ff2) so for the time being we don't need the backport (except for consistency).

@zakkak We should also provide the backport in 24.1. There is a good chance the same will happen there. Lets fix it before it's an issue. The backport is low-risk enough.

zakkak:
Done in https://github.com/graalvm/mandrel-packaging/pull/439

PR #451, opened 2024-09-23
Improves error handling for release script

Karm:
Deals with some errors I used to tackle manually, e.g. missing JDK version and what it means, missing tag etc.

### DIFF
--- a/release-scripts/mandrel-release.java
+++ b/release-scripts/mandrel-release.java
@@ -318,13 +318,22 @@ class GitHubOps
                 Log.error("Please create tag " + tag + " and try again");
             }
 
-            final String changelog = createChangelog(repository, milestone, tags);
+            /*
+             * There used to be more JDK versions per 1 Mandrel release, e.g. Mandrel 21.3 could work with both
+             * Java 11 and 17. We aim at aligning releases with Java releases now, and we could simplify this code as
+             * soon as we completely get rid of all Mandrel 22.3.
+             * Mandrel 25 for JDK 25 will conclude the renaming.
+             */
+            final List<String> jdkVersions = jdkVersionsUsed.stream().filter(v -> v.length() > 2).sorted().toList();
+            final String changelog = createChangelog(repository, milestone, tags, jdkVersions.size() == 1 ? jdkVersions.get(0) : null);
+            // e.g. deals with "23+37" too. Things like 24-beta+16-ea should not happen in a release script.
+            final String jdkVersionForExampleURLs = jdkVersions.get(0).split("[+.]")[0].trim();
 
             if (dryRun)
             {
                 Log.warn("Skipping release due to --dry-run");
                 Log.info("Release body would look like");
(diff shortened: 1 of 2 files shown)


release-scripts/mandrel-release.java
```diff
@@ -318,13 +318,22 @@ void createGHRelease(Set<String> jdkVersionsUsed, boolean linuxUpload, boolean w
                 Log.error("Please create tag " + tag + " and try again");
             }
 
-            final String changelog = createChangelog(repository, milestone, tags);
+            /*
+             * There used to be more JDK versions per 1 Mandrel release, e.g. Mandrel 21.3 could work with both
+             * Java 11 and 17. We aim at aligning releases with Java releases now, and we could simplify this code as
+             * soon as we completely get rid of all Mandrel 22.3.
```
zakkak, line 324: @Karm isn't 22.3 already out of support? If I am not mistaken we currently only support 23.0 (JDK 17) till November, 23.1 (JDK 21), and 24.1 (JDK 23).

release-scripts/mandrel-release.java
```diff
@@ -590,7 +602,17 @@ private String getLatestReleasedTag(List<GHTag> tags)
         if (finalVersions.size() == 1)
         {
             // There is no Mandrel release before that major.minor.micro, return upstream graal tag instead
-            final String upstreamTag = "vm-" + version.majorMinorMicro();
+            final String upstreamTag;
+            if (version.major >= 24 && jdkVersionUsed != null)
```
zakkak, line 606: Do we really need this? `vm-24.1.0` tag is still available, see https://github.com/oracle/graal/tree/vm-24.1.0

Karm: @zakkak Hmm, the script did not find it though, I guess it's because we are not pushing it to our repo?
Anyway, the `jdk-version` tagging scheme is the new thing we are migrating towards...right?

release-scripts/mandrel-release.java
Karm: @zakkak It is out of support. Do you think the idea of 1 Mandrel 2 JDKs is dead then? Meaning, can I go and purge all that logic from the script?

There is a lot of data structures and logic wired to support that we are publishing more JDK builds for that particular Mandrel release.

release-scripts/mandrel-release.java
zakkak: > I guess it's because we are not pushing it to our repo?

Your guess is right :) I just pushed it.

> Anyway, the jdk-version tagging scheme is the new thing we are migrating towards...right?

Good point. That's probably worth asking upstream. What tag scheme do they plan to keep after the versions align with the release of 25?

release-scripts/mandrel-release.java
zakkak: > Do you think the idea of 1 Mandrel 2 JDKs is dead then?

It seems so. In theory there is nothing preventing us from doing it, but there doesn't seem to be any demand for it. Let's discuss it more broadly before doing it though.

PR #481, opened 2025-02-19
[CI] Add support for JDK pinning

jerboaa:
Sometimes it's useful to pin CI to an *older* JDK build. This patch adds
this by allowing to set an ENV variable to the proper GH release to
pin to.


jerboaa:
Test run with "pinned" `JDK-25+8` is here (uses [this commit](https://github.com/jerboaa/mandrel-packaging/commit/e6c4280f4cb45ed8b078dd270a83dcf7a6ec7c53) to pin to it):
https://github.com/jerboaa/mandrel-packaging/actions/runs/13416782255

The run in this PR is the old latest EA JDK build (`JDK-25+9` currently).

zakkak (approved):

PR #489, opened 2025-04-16
Fix mandrel build with JDK 25+18

jerboaa:
This adds the new required dependency module,
`org.graalvm.nativeimage.shaded.capnproto`, which is already built.
Be sure to copy the required jar to the `lib/svm/builder` directory.

Closes: https://github.com/graalvm/mandrel/issues/838


jerboaa:
@Karm @zakkak PTAL. This gets JDK 25 CI going again. The dep is needed for the layered image stuff.

jerboaa:
This is a rather trivial change and we have broken CI for JDK 25 for a while now so I'm going to merge. We can address any review feedback in retrospect.

Issue #494, opened 2025-05-23
Mandrel fails to build latest graal master due to org.graalvm.espresso.shared.svm dependency


zakkak:
Latest builds fail with

```
Error occurred during initialization of boot layer
java.lang.module.FindException: Module org.graalvm.espresso.shared.svm not found, required by org.graalvm.nativeimage.builder
```

Which was introduced by https://github.com/oracle/graal/pull/11235

PR #495, opened 2025-05-23
Include newly added espresso-svm.jar to Mandrel builds

zakkak:
I also experimented with removing the dependency but that required
patching code and not just the suite.py files, so I decided to keep the
dependency. After all, we will most likely want to support Crema (which
depends on this jar) in the future.

Closes: https://github.com/graalvm/mandrel-packaging/issues/494

Note: I pinned the JDK version in the CI pipeline since building with 25+23 depends on https://github.com/oracle/graal/pull/11217 which is not merged yet

### DIFF
--- a/build.java
+++ b/build.java
@@ -876,6 +876,7 @@ class Mx
     final Path compilerDistPath;
     final Path truffleDistPath;
     final Path substrateDistPath;
+    final Path espressoSharedDistPath;
     final Path sdkDistPath;
 
     final String[] mxbuildDir;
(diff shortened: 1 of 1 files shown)


Karm (approved):
Espresso Crema it is. Ack.

zakkak:
> @zakkak Wouldn't
> 
> ```
>   TEMURIN_PINNED_RELEASE: jdk-25%2B22-ea-beta
> ```
> 
> suffer from [adoptium/api.adoptium.net#1482 (comment)](https://github.com/adoptium/api.adoptium.net/issues/1482#issuecomment-2871916436) ?
> 
> Perhaps that's also causing the GHA fail, dunno:
> 
> ```
>     TEMURIN_API_URL_LATEST: https://api.adoptium.net/v3/binary/latest/25/ea
>     TEMURIN_PINNED_RELEASE: jdk-25%2B22-ea-beta
>     TEMURIN_API_URL: https://api.adoptium.net/v3/binary/version/jdk-25%2B22-ea-beta
>     ARCH: x64
> tar: Error opening archive: Unrecognized archive format
> ```

Thanks for the link, however it looks good on AArch64 so I am not sure it's related.

In any case:

1. we don't ship intel-mac builds
2. this is just temporary to see that the fix works, even if there is something wrong with 25+22 on intel-macs it seems irrelevant
3. I don't plan to merge the "pinning"
4. 25+22 is the last known working version

so I am not that worried about it.

PR #496, opened 2025-05-26
MacOS hint xattr

Karm:
fixes https://github.com/graalvm/mandrel/issues/847


zakkak (approved):
I agree, but would also like to investigate how we could avoid the need for this.

